PR #2422, opened 2026-04-08
⚡ Bolt: Optimize redundant norm calculations in collision checker

dieterolson:
### 💡 What
Cached the difference vector (`diff = pb - pa`) and its norm (`norm = np.linalg.norm(diff)`) in the geometric collision distance checks inside `src/robotics/planning/collision/collision_checker.py`. 

### 🎯 Why
During execution of `compute_distance`, the `if np.linalg.norm(pb - pa) > 1e-10:` condition and subsequent `normal = (pb - pa) / np.linalg.norm(pb - pa)` statement were computing the difference between two position vectors and invoking expensive `np.linalg.norm` operations twice per loop iteration. This introduces redundant Python-to-C API overhead inside a high-frequency path.

### 📊 Impact
Halves the number of vector subtractions and `np.linalg.norm()` calls in the narrow-phase geometric distance computations while preserving dimension-agnostic safety. Reduces overhead inside the hot loop.

### 🔬 Measurement
Tests confirm identical functionality without regressions:
1. `pytest tests/unit/robotics/test_planning_collision.py`
2. `ruff check src/robotics/planning/collision/collision_checker.py`

---
*PR created automatically by Jules for task [17593742630026773935](https://jules.google.com/task/17593742630026773935) started by @dieterolson*


dieterolson:

PR #2426, opened 2026-04-08
⚡ Bolt: [performance improvement] Cache redundant vector norm calculations

dieterolson:
💡 What: Cached the difference vector `pb - pa` and its norm in local variables in `CollisionChecker.compute_distance` to avoid calculating the vector difference and `np.linalg.norm` twice.
🎯 Why: `np.linalg.norm` operations are slow and redundant in tight loops that iterate over collision pairs. Reusing the norm reduces overhead.
📊 Impact: Halves the number of `np.linalg.norm` calls when a collision point normal is computed, improving the distance computation loop performance.
🔬 Measurement: Run `pytest tests/unit/robotics/test_planning_collision.py tests/deployment/test_collision.py` to ensure distance computation still outputs the correct normals.

---
*PR created automatically by Jules for task [16107194818799138666](https://jules.google.com/task/16107194818799138666) started by @dieterolson*


dieterolson:

Issue #2493, opened 2026-04-10
Default model and config registries resolve the wrong paths and miss bundled manifests


dieterolson:
## Summary
Several default configuration helpers compute repository paths incorrectly, causing them to miss bundled assets or create duplicate directories under the wrong tree.

## Problems
- `StandardModelManager` computes a suite root under `src/shared` and then appends another `shared`, producing `src/shared/shared/...` paths.
- Both model-registry implementations default to `config/models.yaml`, but the real manifest lives at `src/config/models.yaml`.
- Default construction therefore yields empty registries or duplicate directories instead of loading bundled configuration.

## Impact
Fresh or default-initialized installs can fail to discover bundled models and manifests, causing missing assets and inconsistent runtime state.

## References
- `src/shared/python/config/standard_models.py:31`
- `src/shared/python/config/standard_models.py:35`
- `src/shared/python/config/standard_models.py:36`
- `src/shared/python/config/standard_models.py:37`
- `src/shared/python/config/model_registry.py:36`
- `src/shared/python/config/model_registry.py:81`
- `src/launchers/model_registry.py:28`
- `src/launchers/model_registry.py:82`
- `src/config/models.yaml:1`

PR #2510, opened 2026-04-10
Fix default model registry paths

dieterolson:
Fix the default path resolution cluster for standard models and both model registry implementations.

Changes:
- Default StandardModelManager to the repo's src root so it resolves src/shared/urdf instead of src/shared/shared/urdf.
- Default shared ModelRegistry to the real src/config/models.yaml file.
- Default launcher ModelRegistry to src/config/models.yaml so load(root_path) still resolves against the repo layout.
- Add focused regression tests for all three defaults.

Verification:
- python3 -m py_compile on the 3 touched source files and 3 touched test files
- Direct smoke check for StandardModelManager + shared ModelRegistry path resolution
- Direct smoke check for launcher ModelRegistry default path and load behavior


dieterolson:
Closing as superseded. The default model registry path fixes have already been incorporated into staging via:
- PR #2550 (fix(config): fix double-shared path in StandardModelManager - closes #2493)

The branch is 200+ commits behind staging and cannot be cleanly rebased.

PR #2512, opened 2026-04-10
Fix build and installer paths

dieterolson:
## Summary
- fail packaging early when `ui/dist` is missing instead of silently shipping a broken UI
- make `install.sh` work both from a checkout and from `curl ... | bash`
- point Windows setup/MSI packaging at the real `src/...` asset and config layout
- add focused regression coverage and update `SPEC.md` for the packaging behavior

## Validation
- `/tmp/upstreamdrift-tools/bin/ruff check build_hooks.py setup_golf_suite.py installer/windows/setup.py installer/windows/setup_config.py tests/unit/test_build_hooks.py tests/unit/test_setup_golf_suite.py tests/unit/installer/test_windows_setup.py tests/unit/test_install_script.py`
- `bash -n install.sh`
- `python3 -m compileall build_hooks.py setup_golf_suite.py installer/windows/setup.py installer/windows/setup_config.py tests/unit/test_build_hooks.py tests/unit/test_setup_golf_suite.py tests/unit/installer/test_windows_setup.py tests/unit/test_install_script.py`
- direct build-hook smoke checks for existing `ui/dist` and CI-without-bundle failure
- direct install-script harness checks for remote and local install sources
- direct Windows installer `setup_config` smoke check for `src/...` include paths and entry scripts

## Notes
- `pytest` was hanging in this shell even for single-file runs, so validation here used deterministic direct harness checks plus lint/compile validation.

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -27,7 +27,7 @@
 | **Primary Language(s)** | Python 3.10+, Rust, TypeScript                     |
 | **License**             | MIT                                                |
 | **Current Version**     | 2.1.0                                              |
-| **Spec Version**        | 1.0.74                                             |
+| **Spec Version**        | 1.0.75                                             |
 | **Last Spec Update**    | 2026-04-10                                         |
 
 ## 2. Purpose & Mission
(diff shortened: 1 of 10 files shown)


dieterolson:
Superseded by #2564 which is a clean rebase onto staging. The original PR was based on main and could not be cherry-picked cleanly due to 200+ divergent commits.

PR #2629, opened 2026-04-12
chore: force self-hosted runners to conserve github action minutes

dieterolson:
By bypassing ubuntu-latest and the pick-runner dispatcher, we forcibly route all jobs to self-hosted runners entirely, eliminating GitHub cloud Action minute consumption. This change is entirely reversible via git revert or mass sed replacement.


dieterolson:
Superseded by the fleet/mode-aware-pick-runner approach. The new PRs add a FLEET_RUNNER_MODE org variable toggle (local/cloud/hybrid) instead of hardcoding `self-hosted`.

PR #2630, opened 2026-04-12
chore(temp): harden all 55 workflows to d-sorg-fleet runner (Apr)

dieterolson:
**TEMPORARY until May 1, 2026** — directs all workflow jobs exclusively to the local `d-sorg-fleet` self-hosted runner. No cloud fallback. Jobs will queue if the runner is offline.


dieterolson:
Superseded by the fleet/mode-aware-pick-runner approach.

PR #2723, opened 2026-04-17
[codex] docker runtime defaults to API entrypoint

dieterolson:
## What changed
- Switched the runtime Docker image off the interactive shell default and onto `src.api.server:app` on `0.0.0.0:8001`.
- Documented the runtime container launch behavior in `SPEC.md`.
- Added a regression assertion that the Dockerfile keeps the API entrypoint as the default command.

## Why
A plain container start should serve the API instead of dropping into a shell, which makes the image usable in local runs and safer as a deployment primitive.

## Validation
- `python3 -m pytest tests/test_docker_integration.py -q`

## Notes
- `python3 -m pytest tests/unit/test_start_api_server.py -q` currently fails during collection in this environment because the local Python/Numpy/Pandas stack is inconsistent (`pandas` expects a newer NumPy ABI).


dieterolson:
Closing this stale/broad PR in favor of a fresh narrow salvage item: #2786. This branch is now blocked by repo-wide Ruff formatting churn unrelated to the Docker runtime entrypoint change, and the broader replacement PR #2736 is also dirty/conflicted. Keeping #2786 as the tracking issue gives us a clean slate to reapply only the runtime-entrypoint regression and docs work.

dieterolson:
Converted this broad failing PR into a focused implementation issue: https://github.com/D-sorganization/UpstreamDrift/issues/2787

PR #2755, opened 2026-04-17
⚡ Bolt: Optimize squared distance calculation using np.einsum

dieterolson:
💡 What: Replaced `np.sum(diff ** 2, axis=1)` with `np.einsum("ij,ij->i", diff, diff)` in `golf_visualizer_widget.py`. Added `spec-exempt` label since this is a pure performance optimization without behavior change.
🎯 Why: `np.einsum` avoids intermediate array allocations and is significantly faster when performing reductions over small inner dimensions.
📊 Impact: The max distance calculation is ~2x faster, reducing overhead in visualizer rendering updates.
🔬 Measurement: Ran a standalone perf test that showed `np.einsum` completing in ~0.46s compared to ~0.86s for `np.sum(..., axis=1)` on 10,000 points over 1,000 iterations.

---
*PR created automatically by Jules for task [2542959308789365869](https://jules.google.com/task/2542959308789365869) started by @dieterolson*


dieterolson:
Assessment: the PR diff is narrow (SPEC plus a visualization np.einsum optimization), while the failing Python 3.10 job is broad baseline fallout: API dispatcher validation failures, Dockerfile digest/test expectation mismatch, installer dictionary-iteration error, and cancelled 3.11/3.12 jobs after the 3.10 failure. I am not folding those unrelated baseline repairs into this performance PR; this should either be rebased after the baseline fixes land or have the failing shared checks repaired in a separate CI/baseline PR.

PR #2756, opened 2026-04-17
fix: filter non-level metadata from engine capabilities API response

dieterolson:
Fixes bug where spatial_jacobian_order (value: angular_linear) was being included as a CapabilityLevelResponse entry, causing test_capabilities_have_required_fields to fail. Added _VALID_LEVELS filter to skip non-level metadata keys. Updates SPEC.md.


dieterolson:
Closing this conflicted/stale PR in favor of focused follow-up issue #2797.

The replacement issue captures the actionable change and validation path without the unrelated branch churn.

PR #2793, opened 2026-04-19
perf(docker): slim runtime image — 10.4 GB → 3.18 GB

dieterolson:
## Summary

Ports the slim Docker runtime from Golf_GAAI_Sandbox [b8348af](https://github.com/D-sorganization/Golf_GAAI_Sandbox/commit/b8348af9), with two additional fixes needed to make the API server actually boot inside the resulting image.

**Base flip:** `continuumio/miniconda3:24.11.1-0` → `python:3.12-slim` + pip venv.
**Runtime size:** 10.4 GB → **3.18 GB** (70% reduction, verified locally).
**Size-gate budget:** 16 000 MB → **4 000 MB**.

## What the runtime image still includes

Python 3.12 venv with: `mujoco`, `pinocchio` (pip wheel), `fastapi`, `uvicorn`, `pandas`, `matplotlib`, `sympy`, `defusedxml`, `numpy`, `scipy`, full auth stack. Headless GL libs + ffmpeg for video encoding + curl for healthcheck. Non-root `golfer` user, `/workspace` with `src/`, `pyproject.toml`, entrypoints.

## What the runtime image drops

| Dropped | Why safe |
| --- | --- |
| `pydrake`, `crocoddyl` | Only 2 files import pydrake, both under `src/engines/physics_engines/drake/` which has its own Dockerfile. `engine_availability.py` detects missing imports at runtime and marks engines unavailable — API returns "engine not available" instead of crashing. |
| `myosuite` | Same defensive-import pattern. |
| `ray` | 0 imports in `src/`. Stays in `training` stage. |
| `PyQt6` + X11/XCB + xvfb | GUI-only. API server is headless, verified booting after the dashboard lazy-import fix below. |
| `opencv` (`cv2`) | Only `video` route uses it. That route is now optionally loaded — `route_registry` already skips failing modules and logs a warning. |
| PyTorch + CUDA toolkit | Stays in `training` stage. |

## Fixes to the original slim design

The Golf slim Dockerfile ships an image whose `CMD` actually fails on import. Verified: running `from src.api import server` on the Golf image → `ModuleNotFoundError: No module named 'pandas'`. Their smoke test only imports `mujoco, numpy, scipy, fastapi, pinocchio`, so the failure never surfaces in CI.

Two fixes land here:

1. **Add `pandas`, `matplotlib`, `sympy`, `defusedxml` explicitly.** These used to come transitively from the conda base; `requirements.lock` has never included them. `src/shared/python/data_io/common_utils.py` imports pandas at module top-level, so the entire engine_manager chain needs it.
2. **Make `src/shared/python/dashboard/__init__.py` lazy.** Previously it did `from .launcher import launch_dashboard` and `from .window import UnifiedDashboardWindow` at import time, both of which transitively load PyQt6. `simulation_service` imports `GenericPhysicsRecorder` from `dashboard.recorder`, so the whole API import chain was pulling in Qt. Switched to `__getattr__` lazy loading — GUI consumers see the same public API, headless consumers no longer need PyQt6.

## Verification (local on DeskComputer)

```text
docker build --target runtime -t upstream-drift:slim-verify .   → 3.18 GB
docker run -d -p 18002:8001 upstream-drift:slim-verify
curl http://localhost:18002/health   → {"status":"healthy","engines_available":8,...}
```

Startup logs: 19 of 20 API route modules register. Only `video` is skipped (expected — cv2/mediapipe deliberately absent).

## Functionality and CI impact

**Preserved:** host-side ruff/pytest/mypy CI (runs on runner, not Docker). Engine-specific Dockerfiles. Local developer GUI. `training` stage retains PyTorch + RL stack. Multi-stage build caches unchanged.

**Changed:** `docker-security-scan` now builds `--target runtime` (3 GB) instead of default training (13+ GB). `docker-compose` PATH `/opt/conda/bin` → `/opt/venv/bin`. Workflows that expected `pydrake`/`myosuite`/`cv2` in the runtime image now get the "engine unavailable" path — dedicated engine Dockerfiles remain the supported way to get them.

## Files

- `Dockerfile` — slim two-stage + training.
- `.dockerignore` — excludes Simscape MATLAB models, tests/, conftest.py, build_hooks.py, requirements-dev.lock.
- `docker-compose.yml` — PATH fix.
- `.github/workflows/docker-size-gates.yml` — 4 GB budget, runner routing, extended smoke test.
- `.github/workflows/docker-security-scan.yml` — `--target runtime`.
- `src/shared/python/dashboard/__init__.py` — lazy `__getattr__`.

## Test plan

- [ ] `docker-size-gates` job builds runtime image under 4 GB; smoke imports succeed.
- [ ] `docker-security-scan` Trivy scan completes on the slim image.
- [ ] Host-side pytest + ruff + mypy continue to pass (unaffected — no production logic changes beyond the lazy-import).
- [ ] No existing test depends on eager `from dashboard import launch_dashboard` at import time (lazy `__getattr__` returns the same object, so `dashboard.launch_dashboard(...)` still works).

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -42,6 +42,15 @@ docs/
 *.md
 !README.md
 
+# MATLAB Simscape models — large, not needed in container
+src/engines/Simscape_Multibody_Models/
+
+# Test and dev files not needed in image
+tests/
+conftest.py
+build_hooks.py
+requirements-dev.lock
+
 # Data and logs
 *.db
 *.log
--- a/.github/workflows/docker-security-scan.yml
+++ b/.github/workflows/docker-security-scan.yml
@@ -31,7 +31,8 @@ jobs:
         uses: actions/checkout@v6
 
       - name: Build Docker image
-        run: docker build -t upstreamdrift:ci .
+        # --target runtime avoids pulling in the ~10 GB training stage (PyTorch + CUDA)
+        run: docker build --target runtime -t upstreamdrift:ci .
 
       - name: Run Trivy vulnerability scanner
         uses: aquasecurity/trivy-action@master
--- a/.github/workflows/docker-size-gates.yml
+++ b/.github/workflows/docker-size-gates.yml
@@ -30,7 +30,7 @@ jobs:
           fi
   docker-smoke-and-size:
     needs: pick-runner
-    runs-on: self-hosted
+    runs-on: ${{ needs.pick-runner.outputs.runner }}
     steps:
       - name: Check out repository
         uses: actions/checkout@v6
@@ -51,9 +51,8 @@ jobs:
 
       - name: Verify Image Size Budget
         run: |
-          # Set the maximum allowed size in MB (e.g., 16000 MB = 16GB)
-          # Increased to 16GB to accommodate large dependencies (Drake, MuJoCo, PyTorch)
-          MAX_SIZE_MB=16000
+          # Runtime image budget: slim venv (~1 GB) + src/ models (~650 MB) + base + libs ≈ 4 GB
+          MAX_SIZE_MB=4000
 
           # Get size in bytes
           IMAGE_SIZE_BYTES=$(docker image inspect upstream-drift:runtime --format='{{.Size}}')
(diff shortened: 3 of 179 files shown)


dieterolson:
Closing this PR as part of fleet PR remediation. The slim runtime direction is useful, but the branch has grown to 179 files, changed head SHA, and currently has no check-runs on the latest head. I preserved the findings, attempted fixes, and acceptance criteria in issue #2810 so this can come back as a focused PR.

PR #2933, opened 2026-04-21
fix(launcher): restore synchronous _cleanup_processes for legacy callers

dieterolson:
## Summary

PR #2930 changed `_cleanup_processes` to delegate to the async `_schedule_cleanup()`, which broke tests expecting synchronous cleanup behavior.

This patch restores `_cleanup_processes` to be synchronous: it computes finished keys inline and calls `_on_cleanup_finished` directly. The timer-based path still uses `_schedule_cleanup` for non-blocking async cleanup.

## Test plan

- [ ] `test_cleanup_processes` passes
- [ ] `test_on_cleanup_finished_updates_running_processes` passes (if it exists)
- [ ] `ProcessCleanupWorker` TypeError is gone

Fixes test regression introduced by PR #2930.

### DIFF
--- a/src/launchers/golf_launcher.py
+++ b/src/launchers/golf_launcher.py
@@ -640,9 +640,12 @@ class GolfLauncher(
             self.lbl_status.setStyleSheet(Styles.STATUS_INACTIVE)
 
     def _cleanup_processes(self) -> None:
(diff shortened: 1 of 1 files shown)


dieterolson:
Closing — pushing the fix directly to fix/qrunnable-signals instead.

PR #3316, opened 2026-04-26
fix: correct regex validation of ONLINE var in workflows

dieterolson:
Fixes a bash syntax error where the \ONLINE\ var could contain json.


dieterolson:
Superseded by manual workflow stabilization and kill-switch implementation on main.

PR #3932, opened 2026-05-03
Fix DbC fail-open and vacuous assert True tests

dieterolson:
## Summary

This PR addresses two issues from the github-audit documentation:

1. **Issue 004-dbc-fail-open.md** - Fixed the Design by Contract (DbC) fail-open issue
2. **Issue 003-tdd-coverage-crisis.md** - Removed vacuous `assert True` tests

### DbC Fail-Open Fix (004-dbc-fail-open.md)

- **Add ContractEvaluationError exception class**: New exception type for contract evaluation failures
- **Fix _evaluate_precondition**: Now raises ContractEvaluationError instead of silently returning True when precondition evaluation fails due to signature mismatches or type errors
- **Fix postcondition decorator**: Raises ContractEvaluationError on evaluation failures (ZeroDivisionError, AttributeError, etc.) instead of silently returning
- **Fix import paths**: Updated all contracts-related files to use 'src.shared.python' instead of 'shared.python' for consistent module resolution
- **Export ContractEvaluationError**: Added to main contracts module public API

### Vacuous Test Removal (003-tdd-coverage-crisis.md)

- Removed `assert True` statements from test files that provide no validation value
- Tests that verify 'no exception raised' now document this explicitly without vacuous assertions

**Files modified:**
- tests/unit/engines/drake/test_example.py
- tests/unit/engines/pinocchio/test_example.py

## Testing

Verified that:
1. Normal contract operations still work correctly
2. ContractEvaluationError is raised when preconditions cannot be evaluated (e.g., lambda expects different number of arguments)
3. ContractEvaluationError is raised when postconditions fail evaluation (e.g., division by zero in lambda)

## Impact

**DbC Fix**: This is a **behavior change** - contracts now fail closed rather than silently passing. Code that relied on broken lambdas being silently ignored will now raise ContractEvaluationError.

**Test Cleanup**: No behavioral change - tests continue to pass, but now with meaningful assertions.

## Related Issues

- Fixes: 004-dbc-fail-open.md (DbC precondition evaluator fails open)
- Partially addresses: 003-tdd-coverage-crisis.md (TDD coverage crisis)


dieterolson:
Checking quality-gate failure details...

Issue #4477, opened 2026-05-08
feat(motion-matching): `.mat` club-target loader (TW/GW × ProV1/Wiffle)


dieterolson:
# feat(motion-matching): `.mat` club-target loader (TW/GW × ProV1/Wiffle)

## Why

The motion-matching dispatcher `src/shared/python/motion_matching/load_club_target.py` currently routes `.xlsx` → `loaders/excel.py` and `.c3d` → `loaders/c3d.py`. We also have eight `.mat` files (`TW_ProV1.mat`, `TW_wiffle.mat`, `GW_ProV1.mat`, `GW_wiffle.mat`, plus `_targetKinematics.mat` siblings) under `src/engines/physics_engines/pinocchio/data/<dataset>/` that encode the same swings as the xlsx but with a cleaner schema:

```
data.time            (Nx1)         seconds
data.midhands_xyz    (Nx3)         metres
data.midhands_dircos (Nx3x3)       rotation
data.clubface_xyz    (Nx3)         metres
data.clubface_dircos (Nx3x3)       rotation
params.Address       int           1-based frame index
params.TopOfBackswing int          1-based frame index
params.Impact        int           1-based frame index
params.Finish        int           1-based frame index
params.impact_frame  int           same as Impact
```

These files give us a `ClubTarget` with **stamped impact** rather than a kinematic-peak heuristic, AND with **real 3-DOF clubface orientation** — both improvements over the xlsx path. They should be a first-class loader.

## What to build

`src/shared/python/motion_matching/loaders/matlab_dataset.py`:

```python
def load_club_target_mat(path: Path | str, opts: AlignOptions) -> ClubTarget: ...
```

### Behaviour

1. Open with `scipy.io.loadmat(path, squeeze_me=True, struct_as_record=False)`.
2. Read `data.time`, the four kinematic arrays, and `params.Impact` (1-based → 0-based).
3. Convert `data.midhands_xyz` → `butt`, `data.clubface_xyz` → `clubhead`. (Verify and document the convention; the matcher already uses "Mid-hands" and "Center of club face" — these correspond to butt-end and head respectively for our purposes.)
4. Build per-frame quaternion from `data.clubface_dircos` (rotation matrix → quaternion via existing `loaders/_quaternion.py: rotmat_to_quat`).
5. Sanity-check rotation matrices: `det ≈ +1`, columns orthonormal within 1e-3. Reject otherwise with a clear `ValueError`.
[...]

PR #4490, opened 2026-05-08
feat(motion-matching): .mat club-target loader

dieterolson:
## Summary

Adds `load_club_target_mat` (issue #4477) so the motion-matching dispatcher routes `.mat` motion-capture files alongside `.xlsx` and `.c3d`. Delivers a stamped-impact `ClubTarget` with full 3-DOF clubface orientation — both upgrades over the heuristic-impact xlsx path.

- New: `src/shared/python/motion_matching/loaders/matlab_dataset.py`
- Wired into `load_club_target` dispatcher; re-exported from package & loaders `__init__`
- DbC pre/postconditions: path exists, `opts.sample_rate_hz > 0`, return is `ClubTarget`
- Validates rotation stack: `det == +1` and orthonormal columns to 1e-3
- Stamped impact derived from the `t = 0` row in the source time vector (verified across all four canonical fixtures); falls back to integer `params.Impact` only when the time vector does not span zero

## Implementation notes

The eight canonical files store rotation matrices flat-row-major in `(N, 9)` shape (not `(N, 3, 3)` as the issue speculated) — the loader reshapes either form. The `params.Impact` integer in the canonical files (168, 125, 229, 209) does not index the trajectory rows; it appears to reference a swing-segment subset. The reliable stamped-impact signal in those files is the row where `t = 0`, which matches the max-clubhead-speed frame in every file. The loader prefers that signal and uses `params.Impact` only as a fallback for files whose time vectors do not span zero.

## Measured |v_clubhead|@impact (snapshot baselines)

| Trial      | Speed (m/s) | resampled `impact_idx` |
|------------|-------------|------------------------|
| TW_ProV1   | 51.67       | 251                    |
| TW_wiffle  | 51.70       | 251                    |
| GW_ProV1   | 51.72       | 251                    |
| GW_wiffle  | 50.77       | 251                    |

All four are driver-class swings (not iron); the issue body's iron speculation for the GW files is not borne out by the data. Tests use a single physical-plausibility band `[35, 60] m/s`.

## Test plan

- [x] `pytest tests/unit/motion_matching/test_loaders_matlab.py -n auto --timeout=60` (17/17 pass)
- [x] `ruff check` & `ruff format --check` on changed files
- [x] `scripts/ci/check_file_size_budget.py` passes
- [x] Sister loader tests still pass (`test_loaders_excel.py`, `test_loaders_c3d.py`)

### DIFF
--- a/docs/review_archive/comment_tracking.json
+++ b/docs/review_archive/comment_tracking.json
@@ -41,7 +41,8 @@
     "3206539612",
     "3208650036",
     "3210255790",
-    "3210197994"
+    "3210197994",
+    "3210149886"
   ],
   "created_issues": {
     "3197917606": "https://github.com/D-sorganization/UpstreamDrift/issues/4140",
--- a/docs/review_archive/review_comments_2026-05-08.md
+++ b/docs/review_archive/review_comments_2026-05-08.md
@@ -34,3 +34,18 @@ Importing `matplotlib` at module scope makes this test file fail during collecti
 
 ---
 
+### PR #4490: src/shared/python/motion_matching/loaders/matlab_dataset.py:150
+
+Actionable: Yes
+Has Suggestion: No
+
+```
+**<sub><sub>![P1 Badge](https://img.shields.io/badge/P1-orange?style=flat)</sub></sub>  Replace params.Impact fallback for non-zero-centered time**
+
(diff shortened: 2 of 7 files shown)


dieterolson:
Rebased on latest `main`; CI re-running.

PR #5866, opened 2026-05-21
[Architecture] Replace UI Mixins with Composition

dieterolson:
Resolves #5859


dieterolson:
Consolidated into #5874

Issue #8699, opened 2026-08-16
EPIC: BunkerShot3D visualization suite — sole load fields, 3-D shot animation, and sand/ball cross-section analysis


dieterolson:
BunkerShot3D has **no visualization at all** today. `postproc/` contains one module (`wrench_trace.py`), and nothing in the package renders anything — the figures in the recent demo were drawn by a throwaway script against `results.json`, not by the library.

This epic builds the visualization suite. It is scoped around a hard architectural constraint that has to be stated first, because it determines what is buildable and what is not.

## The constraint: F0 does not know what the sand is doing

The requested analyses split cleanly in two, and the split is not a matter of effort.

**3-D Resistive Force Theory is a constitutive shortcut.** F0 integrates an empirical resistive stress over the intruder's wetted surface. It never solves the sand's motion. `IntrusionState` carries clubhead velocity, angular velocity, reference point and free-surface height; `SolverResult` carries `depth_force_n` and `inertial_force_n` **per surface element of the sole**. There is no sand velocity anywhere in the model, because the model never computes one.

So:

| Question | Tier that can answer it |
|---|---|
| Which part of the sole carries the load, and when? | **F0 — available now** |
| How does the contact patch migrate through impact? | **F0 — available now** |
| Where does the head lose speed, and to what? | **F0 — available now** |
| Does sand velocity near the face change through the impact zone? | needs a sand-solving tier |
| What is the sand velocity striking the ball, by region? | needs a sand-solving tier |
| Cross-sections of the sand/impact zone | needs a sand-solving tier |

The second group cannot be obtained from F0 by post-processing, interpolation, or any amount of plotting. The quantity does not exist. Any visualization that appeared to show it would be fabricating a field from a model that has none — which is worse than not having the feature.

There is a further wrinkle on the ball. At F0 the ball is driven by a momentum-transfer model (`compute_bunker_launch`, derived from solver impulse in #8657). Nothing strikes it. "Velocity of the sand impacting the ball" has no referent at F0 even in principle.

[...]

Issue #8698, opened 2026-08-16
Declared sole_camber_area_m2 is silently clamped, confounding any design sweep over camber, width or bounce


dieterolson:
A design sweep over `sole_camber_area_m2` silently sweeps a different variable than the one it declares. Found while building a demo sweep for the epic: the clamp fired on **40 of 77 design points**, moving effective camber across **24.5–61.6 mm²** while the declared value was a constant **48 mm²**.

## The clamp itself is correct — that is not the bug

`geometry/lofting.py:112-119`:

```python
scaled_camber = geometry.sole_camber_area_m2 * (1.0 - relief) ** 2
low, high = constructible_camber_range_m2(...)
# A narrower sole admits a narrower camber band; clamping keeps the
# relieved section as close to the declared camber as its width can
# support.
camber_m2 = min(max(scaled_camber, low), high)
```

This is physically right. A narrow sole geometrically cannot host an arbitrarily large camber area, and silently producing an inconstructible section would be worse. Do **not** remove the clamp.

## The bug is that it is unobservable

A caller declares 48 mm², gets something else, and has no supported way to find out:

- `constructible_camber_range_m2` is **not re-exported** from `bunkershot3d.geometry` or the top-level package. It is reachable only by importing the private-ish `geometry.lofting` module directly, so there is no public way to ask "is my declared camber constructible at this width?" before running.
- No result object carries the effective value back. `DeliveredGeometry` does not mention camber, and neither does the shot result. The clamped value exists only inside the mesher.

So the parameter that reaches the mesh is not the parameter the study declared, and nothing anywhere reports the substitution.

## Why this matters more than a normal papercut

It silently confounds design-of-experiments work, which is the entire point of the `study/` subpackage (`DesignSpace`, `MorrisDesign`, `SaltelliDesign`, `SobolIndices`).

In the demo sweep, every apparent "bounce" trend is really a bounce+camber trend, because camber moved as a hidden function of bounce and sole width. A Morris or Sobol run would confidently attribute variance to a factor the user believes they held fixed — and the sensitivity indices would be wrong in a way no diagnostic would reveal. That is the failure mode sensitivity analysis exists to prevent.

It also violates the repo's own DbC posture: a precondition is being repaired in silence rather than reported.

## Suggested fix

Any one of these closes it; the first two are complementary:

[...]

PR #8728, opened 2026-08-16
feat(bunkershot3d): per-element sole load field and contact-patch evolution (#8705, #8707)

### LINKED ISSUE #8705: A1: Paint per-element sole load onto the mesh — the field F0 already computes and currently discards
`SolverResult` carries `depth_force_n` and `inertial_force_n` **per surface element** of the sole, plus `n_active_elements`. Today all of it is collapsed to a scalar wrench and thrown away.

Render it: the sole mesh coloured by load, animated across the shot.

## Why this is the highest-value view in the epic

It answers the question a wedge designer actually asks — *which part of this grind is doing the work, and when*. Heel-side? Trailing edge? Does the load migrate toward the leading edge as the head decelerates? That is the difference between a grind decision and a guess, and the data is already sitting in the result object.

It also needs **no new physics** — this is F0 data, available at ~17 ms/shot.

## Scope

- Per-element scalar field(s) over the existing sole mesh, animated over time.
- Separate depth-dependent from inertial contribution, since they peak at different moments and a designer wants to know which one a feature is fighting.
- Render through the ADR-0027 viewport layer (`src/shared/python/visualization/viewport.py`), not a bespoke renderer.
- Degrade to a matplotlib still when no 3-D provider is installed (`pyvista`, `vtk`, `plotly` are all absent on the dev machine; `matplotlib 3.10.9` is present).
- Validity status and fidelity tier rendered **in the frame**.

Blocked on the metrics/trace composition defect, since the field needs a full usable trace.

Part of #8699.

### LINKED ISSUE #8707: A3: Animate contact-patch evolution through the shot
The demo found the initial contact patch shrinking **6.80 -> 3.46 cm²** across the bounce sweep, and that shrinkage is the mechanism behind the counterintuitive result that more bounce produced *more* depth and force rather than less.

A static initial-patch number was enough to spot the mechanism. Animating the patch through the shot would explain it.

## Scope

- Contact patch (engaged element set) rendered on the sole and on the sand surface, over time.
- Patch area as a time series, linked to the A4 panel.
- Overlay of where the patch sits relative to the leading edge, since "how quickly does load reach the leading edge" is the practical dig question.

## Caveat to carry

The bounce trend this would illustrate is currently confounded by the silent camber clamp (#8698) — effective camber moved 24.5-61.6 mm² against a declared constant 48 mm² on 40 of 77 points. Fix or surface that first, or this view will animate a bounce+camber effect while labelled as bounce.

Part of #8699.

dieterolson:
Closes #8705. Closes #8707. Part of epic #8699.

`simulate_shot` builds the club wrench as an integral of a per-element
traction and then keeps only the resultant. The workbench recovered part of
that — a 12×12 impulse-density map, summed over the whole strike — and
discarded the rest. This PR makes the field itself a first-class result:
per element, per sample, with the two DRFT terms separated, animated, and
with the engaged element set followed against the leading edge.

Both issues land together because they are the same data on the same surface.

## What is new

**`field.py`** (headless, no Qt, no matplotlib)

- `SoleLoadField` — `(T, E)` load with `depth_normal_force_N` and
  `inertial_normal_force_N` carried separately, plus the verdict and the
  fidelity tier so a frame can be stamped without a caption.
- `LoadScale` — a colour scale fixed over every frame and mergeable across
  designs, per component.
- `ContactPatch` — the engaged set, its area, centroid, and gap to the
  leading edge, over time. Refuses to exist when nothing ever engaged.

**`render.py`** (headless matplotlib) — four panels: the depth field, the
inertial field, the engaged patch against the leading edge, and the patch-area
time series with each term's resultant. `viewport_fallback()` asks the
ADR-0027 viewport layer for a 3-D provider and reports what it got.

**`widgets.SoleLoadFieldWidget`** — transport only (frame, slider, timer, play
button). The workbench results panel becomes two tabs: the existing summed
maps, and the new per-element field, A and B side by side.

**`report.sole_field_text`** — the same numbers in text, so they can be pasted
into an issue and diffed between runs.

## What was reused rather than rebuilt

- `metrics/bounce_map.py` already validates and consumes a per-element sole
  load trace — `SoleLoadField.load_trace()` produces exactly that array, so
  the existing metric is unchanged and gets the field for free.
- `bridge.sole_load_trace` becomes a reduction of the new
  `bridge.sole_load_field`. One replay of the recorded poses now feeds the
  field, the patch and the 12×12 map, so the extra fidelity costs no extra
  solver work.
- `widgets.GridMapWidget` keeps painting both summed maps; it gained an
  optional pinned limit pair rather than being replaced.
- The Qt/matplotlib canvas is the repo's shared `MplCanvas`.
- `SoleLoadMap.density_pa_s` and `sole_map_text` are untouched — this raises
  the load view, it does not remove the existing one.

## Honesty

- **Status and tier render in the frame.** `frame_stamp()` is drawn inside all
  four axes in the status colour. A screenshot of one panel keeps its
  contents and loses its surroundings, so a caption would not survive.
- **Colour scaling is consistent across frames and across designs.** Nothing
  calls a per-frame autoscale; a comparison hands both views one set of
  scales and both bounce maps one ramp.
- **Units on everything**: mm on the sole axes, Pa on the colour bars, cm² on
  the patch area, ms on the time axis, N in the resultant legend.
- **The terms are summed and then clamped, never clamped and then summed.**
  One traction can point outward on a steeply raked element while the
  resultant is still compressive, so clamping each term would invent load.
- **A refused shot carries no field and no patch**, enforced in
  `ShotOutcome.__post_init__`, and clearing the view stops playback.
- Every contract check is a `raise`, never an `assert`.

## Measured on the nominal 58° design (coarse test resolution)

| | |
|---|---|
| Resolution | 104 sole elements × 52 samples |
| Peak depth-term resultant | 1.654 N at 3.75 ms |
| Peak inertial-term resultant | 806.7 N at 3.75 ms |
| Inertial share of the sole's resultant | 99.8 % |
| Contact patch at first contact | 1.72 cm² at 1.00 ms |
| Largest contact patch | 16.97 cm² at 3.75 ms |
| Closest approach to the leading edge | 5.63 mm at 2.25 ms (from 13.7 mm) |

Two findings worth reading:

- On this delivery the two terms do **not** separate in time — both peak at
  3.75 ms — but they separate clearly in **space**: the depth term loads
  15–22 mm behind the leading edge and the inertial term 22–28 mm behind it.
  `peak_time_s()` reports the moment per term rather than asserting a
  difference, because the coincidence is a finding too.
- The 99.8 % inertial share on the sole is well above the ≈0.9 whole-body
  share ADR-0032 predicts at 25 m/s. Under one shared colour ramp the depth
  panel would be uniformly blank, which is why each component keeps its own
  scale with the peak stated in Pa in the frame.

## Defect found in the existing workbench

`GridMapWidget` stretched every grid to its own extremes, so the A and B
bounce-utilisation maps of a comparison were drawn on **different ramps**: two
grinds differing by a factor of two in peak impulse density painted
identically, and the panel a designer reads to choose between them was
systematically uninformative. Fixed by an optional explicit limit pair
computed once over both designs. Without limits the old behaviour is
unchanged, so the single-design path and its tests are untouched.

## Correction, and an API inconsistency it exposed

An earlier revision of this description claimed the default `sm9_58_m` preset
"declares 48 mm², realises 41.39". **That was wrong**, and the review caught
it. The preset declares **42.00 mm²** and sits inside its constructible band,
so the aggregate is not substituted at all. The 48 mm² figure belongs to the
*demo harness's* design, which is a different design.

Measured on the shipped preset via `loft_wedge(g, camber_fit=NEAREST)`:

```
declared 42.00 mm2 | effective 42.00 mm2 | camber_was_clamped False
band (38.70, 42.44) | clamped_stations: 3
```

Both lines are true, and together they are the defect. No aggregate
substitution occurred, yet **three stations along the sole were refitted** to
their local bands. `camber_was_clamped` compared the declared area against the
band the **declared** sole width admits; heel and toe relief narrows the sole
toward the ends, a narrower sole admits a narrower camber band, and so the
relieved stations get refitted while the declaration itself is honoured.

That is #8698's own failure mode — silent substitution invisible to a caller —
reappearing at finer grain, through the *simpler* check that most callers will
reach for. It is not confined to a contrived design; at shipped resolution
**three of the six shipped presets** hit it:

| preset | declared | aggregate flag | stations refitted |
|---|---|---|---|
| `sm9_58_m` | 42.00 mm² (in band) | `False` | **3 of 17** |
| `acushnet_example_1` | in band | `False` | **5 of 17** |
| `tour_shaved_heel_lob` | in band | `False` | **13 of 17** |

### Resolution: renamed, not redefined

Of the two options offered, this takes the **rename**.

Redefining `camber_was_clamped` in place to mean "any station" would silently
change the answer every existing caller computes, under an unchanged name —
which is the same failure mode one level up, and precisely what #8698 exists
to prevent. The rename breaks loudly instead: a stale caller gets an
`AttributeError`, not a plausible wrong boolean. The old names are **not**
kept as aliases, because an alias would leave the misleading answer reachable.

Since both scopes turn out to be genuinely needed, one flag could not serve
either way:

- **`aggregate_camber_was_clamped`** — the declared-versus-effective question,
  unambiguously scoped by its name. The workbench's camber-area line needs
  exactly this, because it prints the declared number.
- **`any_camber_was_clamped`** — aggregate **or** any station. This is the
  honest answer to "is the sole I got the sole I asked for", and it cannot
  read `False` while `clamped_stations` holds anything. A caller reaching for
  one boolean now gets a correct one, which a docstring note alone would not
  have achieved.
- `camber_substitution_m2` → **`aggregate_camber_substitution_m2`**, which
  also ends its name collision with `StationCamber`'s per-station property.

Call sites updated: `bridge.HeadBuild`, `model.DesignEvaluation` (which now
carries the per-station account, since the effective area alone cannot answer
the question), and both uses in `report.py`. `study/` has no call sites. The
camber-area line additionally names the refitted station count, so an in-band
declaration over refitted stations no longer renders as a clean `42.0 mm^2`.

### `PATCH_CONFOUND_CAVEAT` rebased onto the stations

The caveat was gated on the aggregate boolean and worded as though the
*declared* area had been inconstructible. Both are now fixed: it is gated on
`clamped_stations` being non-empty and restated in terms of stations, with the
count printed beside it (`3 of 17 spanwise stations refitted`).

**The caveat stays.** Gated on the aggregate it would have been silent on the
default design; gated on the stations it fires there — which is when a caveat
earns its place. #8707's bounce mechanism still cannot be read off a sweep
that moved camber at the same time.

## Tests

TDD throughout: each of the four increments landed a red test first.

| | before | after |
|---|---|---|
| `tests/unit/tools/bunker_shot_gui` | 145 | 226 |
| `tests/tools/bunker_shot_gui` | 45 | 73 |
| **total** | **190** | **299** |

The scope fix added 17 more on top of that: `test_camber_reporting_8698.py`
4 → 6, `test_sole_field_report.py` 17 → 21, and
`tests/bunkershot3d/geometry/test_wedge_camber_observability_8698.py`
31 → 42. The regression assertion is that the broad flag cannot read `False`
while stations were clamped, pinned directly, over every shipped preset, and
on `sm9_58_m` by number (42.00 mm², in band, 3 of 17). Two further tests pin
that the aggregate flag kept its narrow meaning and that the old names are
gone rather than aliased.

All pass. No existing test was modified except
`test_no_fake_physics`'s no-Qt guard, which was **extended** from
`design, model, report` to also cover `field` and `render`.
`tests/bunkershot3d/metrics/test_bounce_map.py` and
`test_shot_metrics_composition.py` still pass unchanged. `mypy` and the full
pre-commit/pre-push hook set pass; no `--no-verify` anywhere.

## Dependencies

None added. matplotlib 3.10.9 and PyQt6 only; pyvista/vtk/plotly remain
absent, which is exactly why the ADR-0027 selection degrades and says so.


dieterolson:
## Correction to the measured table

The table above quoted the coarse **test** mesh. At the **shipped** 11x24
discretization the numbers, and one conclusion, differ:

| | coarse (5 stations) | shipped (11 stations) |
|---|---|---|
| Resolution | 104 elements x 52 samples | 500 elements x 53 samples |
| Peak depth-term resultant | 1.654 N at 3.75 ms | 1.991 N at 4.00 ms |
| Peak inertial-term resultant | 806.7 N at 3.75 ms | 725.3 N at 4.25 ms |
| Inertial share | 99.8 % | 99.7 % |
| Patch at first contact | 1.72 cm² | 1.309 cm² |
| Largest patch | 16.97 cm² | 16.21 cm² |
| Closest approach to leading edge | 5.63 mm (from 13.7 mm) | 3.39 mm (from 13.09 mm) |

At the shipped mesh the two terms **do** separate in time — depth at 4.00 ms,
inertial at 4.25 ms — and the spatial separation sharpens to 5–18 mm against
20–28 mm behind the leading edge. The one-sample temporal separation does not
survive coarsening, which is exactly why `peak_time_s()` reports the moment per
term rather than the implementation asserting a difference. SPEC.md quotes the
shipped configuration.

dieterolson:
## `phantom-guard` failure is a false positive (guard bug, not this PR)

`quality-gate` — the required check — passes on both runs. `phantom-guard`
fails claiming:

> the diff touches **no files under `src/`, `rust_core/`, or `api/`**

This PR touches **eight** files under `src/`:

```
src/tools/bunker_shot_gui/__init__.py   bridge.py   field.py   gui.py
                          model.py      render.py   report.py  widgets.py
```

The guard's own log shows why it got the wrong answer:

```
fatal: 8afce3f5...0b7463e0: no merge base
num_changed=0
##[notice]0 changed files reported; waiting 10s and retrying (attempt 1/3)
##[notice]API reports 14 changed files; deferring to API count to avoid race-induced false positive
post-retry num_changed=14
##[error]... the diff touches no files under `src/` ...
```

Its shallow clone cannot resolve a merge base for the base SHA it was handed,
so `git diff --name-only` returns nothing. The guard **detects** this and falls
back to the GitHub API for the *count* (0 → 14), but the `src/`-touching test
still runs against the empty **git** file list. So the fallback fixes the
symptom it checks and not the one it acts on.

Locally `git merge-base origin/main HEAD` resolves fine (`60c0adf49`), and
`8afce3f5...` is not an object this repo has — the base SHA the guard was given
is not reachable from `main`.

Not fixed here: it is a workflow defect affecting any PR that hits the race,
not something this branch introduced.

dieterolson:
## Correction on the camber caveat — and a real API inconsistency underneath it

The PR reports that the default preset *"silently clamps camber — `sm9_58_m` declares 48 mm², realises 41.39 mm²"*. I checked that and the numbers do not hold as stated:

```
  declared  : 42.00 mm2
  effective : 42.00 mm2
  clamped   : False
  band      : (38.70, 42.44)
  clamped stations: 3
```

`sm9_58_m` declares **42.00** mm², not 48, and it sits inside its constructible band, so the aggregate is **not** clamped. The GUI does not override it either — `design.py`'s `camber_area_mm2` defaults to `None` and resolves to the preset value. (48 mm² is the *demo harness's* declared camber, which is where that figure comes from; the demo is a different design.)

**But the last line is the interesting one.** `camber_was_clamped` is `False` while `clamped_stations` has **3 entries**. Both are true: no aggregate substitution occurred, yet three individual stations along the sole were fitted to their local band.

That means the caveat this PR adds is pointing at something real, just not at what it says. And it exposes an inconsistency in the #8698 API worth fixing:

- A caller who checks the documented boolean (`camber_was_clamped`) concludes nothing was substituted.
- A caller who inspects `clamped_stations` sees three substitutions.

The whole point of #8698 was that silent substitution is invisible to callers. A boolean that reads `False` while per-station substitution happened reproduces that failure at a finer grain — and it is the *more* likely check, being the simpler one.

Suggested resolution, for whoever picks it up:
- Either make `camber_was_clamped` mean "any substitution occurred at any station" (and add a separate aggregate-level flag if the current meaning is still wanted), or
- rename it so the scope is unambiguous (`aggregate_camber_was_clamped`) and document that `clamped_stations` must also be consulted.

For this PR, the fix is small: base `PATCH_CONFOUND_CAVEAT` on `clamped_stations` being non-empty rather than on the boolean, and restate the caveat in terms of stations. The caveat should stay — it fires on the default design, which is exactly when a caveat earns its place.

Everything else in the PR verified as described.

Issue #2806, opened 2026-04-19
Rebuild markdown-only docs CI trigger fix from stale PR #2719


dieterolson:
## Background
PR #2719 attempted to ensure CI runs for markdown-only docs PRs, but it is conflicted (`mergeable_state: dirty`) and includes broad unrelated changes across 186 files.

Original PR: #2719
Original branch: `codex/unblock-upstreamdrift-doc-checks`

## Intended outcome
Adjust CI trigger/path filtering so markdown-only documentation PRs receive the required checks.

## Implementation notes
- Focus only on the relevant GitHub Actions workflow path filters and required-check behavior.
- Confirm whether docs-only PRs should run the full quality gate or a lightweight docs check that satisfies branch protection.
- Avoid unrelated source/model/archive changes.

## Validation
- Open or simulate a docs-only PR and verify required checks are created.
- Validate workflow syntax.

dieterolson:
Fixed and merged via PR #2816. The docs-only CI workflow now satisfies branch protection requirements.

PR #8747, opened 2026-08-17
fix(ci): stop three workflows publishing the required `quality-gate` check

dieterolson:
## The bug

Branch protection matches required status checks **by context name**. Three jobs were named `quality-gate`, so all three published check runs under the one required context and protection was satisfied by whichever reported:

| Workflow | Job | What it actually checks |
| --- | --- | --- |
| `ci-standard.yml` | `quality-gate` | the real aggregate over 6 gates |
| `quality-gate.yml` | `quality-gate` | repo-wide Law-of-Demeter lint |
| `docs-ci.yml` | `quality-gate` | a docs check |

This is not theoretical. On #8728 (head `20cd114`) all three reported:

- **CI Standard → `failure`**
- Docs CI → `success`
- Quality Gate → `success`

**The PR merged.** The heavy aggregate is also the slowest and most likely to be missing, so the weakest check tends to be the one present.

## Why the names couldn't just be renamed

The duplication was load-bearing. `ci-standard.yml` `paths-ignore`d `docs/**`, `**.md`, `LICENSE`, `.gitignore`; `docs-ci.yml` `paths`-included *exactly those*. The triggers were exact complements — precisely one ran per PR — and naming both `quality-gate` is what guaranteed the required context always reported (#2806).

A required context that never reports blocks a PR **forever**, so renaming without touching the triggers deadlocks in both directions:

| PR touches | CI Standard | Docs CI |
| --- | --- | --- |
| code only | runs | **skipped** |
| docs only | **skipped** | runs |

## The fix — move the skip inside the workflow

- **`ci-standard.yml`** drops `paths-ignore` and always runs. A new `changed-paths` job classifies the diff; every substantive job is gated on `needs.changed-paths.outputs.code == 'true'`. Docs-only PRs cost a dispatcher job plus one diff, not the full suite.
- The **`quality-gate` aggregate** additionally needs `pick-runner` and `changed-paths`, and accepts `skipped` gates *only* on docs-only changes. It fails outright if either prerequisite failed, so an infrastructure failure that skips everything cannot read as a pass.
- **`quality-gate.yml`** → `lod-quality-gate`. No path filter, runs on every PR, safe to require directly.
- **`docs-ci.yml`** → `docs-quality-gate`. **Must not be required** — it only runs on docs-touching PRs, so requiring it would block every code-only PR.

`quality-gate` stays on the CI Standard aggregate, which is what the org ruleset and the bots resolve by exact name.

## Branch protection — coordinated change

> [!WARNING]
> The required context is **not** repo-local. It comes from the **organization** ruleset `Repository_Protections` (id `16717249`, `~DEFAULT_BRANCH`, `current_user_can_bypass: never`), which is also active on Tools, Runner_Dashboard and AffineDrift. **Adding the new contexts there would block merges in every org repo**, since those jobs exist only here.

So the org ruleset stays untouched (`quality-gate` keeps working there, unchanged), and the additive requirement goes on the **repo-scoped** ruleset `Protect Main` (id `11189578`, which currently has no status-check rule):

```
quality-gate      (CI Standard aggregate - always reports)
lod-quality-gate  (LOD gate - no path filter, always reports)
```

This is **not applied yet** — it lands with the merge, not before.

Note: the classic `branches/main/protection` REST endpoint is returning 503 org-wide at the time of writing, so the classic contexts list could not be read or written. The ruleset endpoints respond normally.

## Bots

`Jules-Auto-Repair`, `Jules-PR-AutoFix` and `Bot-CI-Trigger` resolve the literal string `quality-gate`, which is preserved. They select it with `head -1`, so collapsing to a single publisher also stops them reading an arbitrary one of three.

## Tests

- **New:** exactly one job may be named `quality-gate` (regression guard for this bug).
- **New:** `ci-standard.yml` has no `pull_request` path filter.
- Existing workflow-structure tests updated to the new contract.
- Two assertions in `test_docs_ci_governance.py` were silently unenforced — a relative path raised `FileNotFoundError` under the suite's chdir. Now anchored to the repo root.
- Corrected a stale assertion that `quality-gate.yml`'s `runs-on` is a bare `d-sorg-fleet-docker`; it has been the dispatcher expression for some time, so **that test was already failing on main**.

Local run of both affected files: baseline at `main` = 4 failed / 77 passed; this branch = 1 failed / 82 passed. The one remaining failure (`test_unit_gate_sparse_checks_out_pinned_tools_for_ownership_guard`) fails identically on `main` and is unrelated (Tools vendor pin).

Opened as a draft deliberately, to hold it against auto-merge automation until the ruleset change is applied in the same window.


dieterolson:
## Audit: the blast radius is the entire recent merge history, not a few PRs

I sampled the 30 most recently merged PRs and measured the **`quality-gate` job** inside `ci-standard.yml` — not the workflow conclusion, which can fail for unrelated reasons.

| Outcome | Count |
|---|---|
| `CI Standard` ran, `quality-gate` job = **failure** | **26 of 26** |
| `CI Standard` never ran (docs-only, `paths-ignore`) | 3 |
| still in progress | 1 |

Every single merged PR whose aggregate actually ran had that aggregate **failing**. Not a minority — all of them.

### Mechanism, confirmed on #8723

Three check-runs named `quality-gate` on that head SHA:

| Workflow | `quality-gate` |
|---|---|
| `ci-standard.yml` | **failure** |
| `quality-gate.yml` | success |
| `docs-ci.yml` | success |

It merged. This is precisely what this PR describes, and #8728 was not an isolated case.

### What the aggregate was failing on

The sub-jobs are consistent across the sample: `repo-structure-gates`, `shared-tools-consumer-contract`, `unit-test-gate`, `code-quality`, `dependency-consistency`, `alembic-postgres-round-trip`, and `tests (3.11)` / `tests (3.12)`.

So this is not a flaky-infrastructure story. `main` has carried real, repeated gate failures that branch protection was reporting as satisfied.

### The docs-only hole is separate and also real

Three PRs merged with **no** `CI Standard` run at all — `ci-standard.yml` has `paths-ignore: ["docs/**", "**.md", ...]`. For those, `quality-gate` came only from `docs-ci.yml`. Renaming contexts without addressing that will make docs PRs unmergeable rather than under-verified, which is presumably why the duplication was load-bearing in the first place.

### Self-implicating note

#8723 is mine. I shepherded it to merge on a green `quality-gate` and reported it as green. It was green — from `Docs CI` and `Quality Gate`, while the real aggregate was red. I had no way to see that from the PR view, which is the point.

### Suggested follow-up once this lands

Whatever `main` is currently failing should be triaged as a batch rather than per-PR, since it accumulated invisibly. Tracked alongside #8754, which is blocking this PR from merging at all.

Issue #8766, opened 2026-08-18
Unit-test-gate debt ledger: 465 quarantined tests (hollow-merge era)


dieterolson:
# Unit-test-gate debt ledger: 465 quarantined tests (hollow-merge era)

## What this is

`scripts/config/unit_gate_quarantine.json` skips 465 test node IDs in the Green-Suite Unit Gate (env-gated by `UNIT_GATE_QUARANTINE=1`, set only in that CI step). Every one of them fails **deterministically** on current `main` — verified identical across two independent CI runs (32186948436, 32188431508, 2026-08-18). They are debt, not flakes.

## Why a ledger instead of fixes

These tests accumulated while the required `quality-gate` context was satisfied by lightweight workflows (the context collision fixed by #8747) — 26+ PRs merged with the real aggregate failing (see the audit on #8747), and #8662 gave suite markers to 365 tests that had **never run in the gate**, unmasking months of older drift at once. The failure surface spans ~115 files and multiple subsystems; fixing it inline would keep the only honest gate red indefinitely, which is exactly the state that produced this debt. This mirrors the sanctioned DRY-quarantine ledger pattern (#8763): record the debt explicitly, keep the gate meaningful for new regressions, burn the ledger down.

**The ratchet rule: entries may only be removed, never added.** A new failure that is not in the ledger reds the gate and blocks merge, as it should.

## Root-cause clusters (burn-down guidance)

| Cluster | ~Tests | Signature | Notes |
| --- | --- | --- | --- |
| Theme API drift | 53 | `cannot import name 'DARK_THEME' / 'get_current_colors' from 'shared.python.theme'` | Shadow-tree theme no longer exports names tests expect |
| UIBuildHook drift | 29 | `TypeError: UIBuildHook() takes no arguments` | `tests/unit/test_build_hooks.py` vs production constructor |
| Sidekick ownership manifest | 28 | `RuntimeError: Sidekick ownership manifest must resolve every entry to UpstreamDrift: chat/` | Manifest/resolution contract broken for `chat/` |
| Import-alias parent-attr trap | ~22 | `AttributeError: module 'src.shared.python' has no attribute 'config'` | `monkeypatch.setattr` dotted-string walk hits the aliased parent object; fix pattern: import the module object directly and `setattr` on it (see #8762 for the same class of fix) |
[...]

Issue #8754, opened 2026-08-18
CI: Pull-request workflows stall in action_required, so no PR can satisfy the required quality-gate


dieterolson:
## Symptom

No pull request in this repository can currently merge. Four open PRs are `BLOCKED` with **zero** status checks reported:

| PR | State | Checks |
|---|---|---|
| #8690 | BLOCKED | 0 |
| #8691 | BLOCKED | 0 |
| #8692 | BLOCKED | 0 |
| #8747 | BLOCKED | 0 |

Over the last 30 workflow runs in this repo, **19 are `action_required`** and the rest are in progress. For comparison, `D-sorganization/Tools` over the same window has zero `action_required`.

## Mechanism

1. `pull_request` workflow runs are created in `action_required` — they need manual approval before executing.
2. Branch protection on `main` requires the `quality-gate` context (`required_status_checks.contexts: ["quality-gate"]`, `strict: true`).
3. An unapproved run never executes, so `quality-gate` never reports.
4. `mergeStateStatus` stays `BLOCKED` indefinitely. Nothing fails — nothing runs. The PR just sits there.

This is silent: the PR shows no red, so it looks like CI is merely slow.

## Second-order problem

Once approved, runs on this branch came back **cancelled** rather than executing, and had to be explicitly re-run. So approval alone is not sufficient to unblock a PR.

## Why this is urgent

**#8747 is one of the blocked PRs.** That PR fixes a real branch-protection hole — three workflows (`ci-standard.yml`, `quality-gate.yml`, `docs-ci.yml`) all publish the required `quality-gate` context, so whichever reports first satisfies protection. Its description shows #8728 merging while `CI Standard` reported **failure**, because two weaker checks reported success.

So the fix for the protection hole is blocked by the CI freeze, while the hole stays open.

## Manual unblock (what worked on #8750)

```bash
# 1. approve the pending runs on the branch
gh run list --repo D-sorganization/UpstreamDrift --branch <branch> --limit 20 \
  --json databaseId,conclusion \
  --jq '.[]|select(.conclusion=="action_required")|.databaseId' \
  | xargs -I{} gh api -X POST repos/D-sorganization/UpstreamDrift/actions/runs/{}/approve

# 2. re-run anything that came back cancelled
gh run rerun <run-id> --repo D-sorganization/UpstreamDrift
```

That took #8750 from `BLOCKED`/0-checks to both `quality-gate` contexts green and merged.

## Decision needed (repo owner)

[...]

PR #8768, opened 2026-08-19
Fix: Clear the 82 Ruff Errors, Bandit Findings, and Frontend Lock Breakage Redding Main's Own CI

dieterolson:
## What

Main's own CI Standard runs are red on **code-quality**, **security-scans**, and **frontend-tests** — debt shipped by hollow merges while the `quality-gate` context collision (#8747 / #8754) let PRs merge with the real aggregate failing. This PR restores all three jobs:

### code-quality — ruff now passes repo-wide (was: 82 errors)
- **Real fixes**: two `PLE0604` `__all__` bugs (lazy-export unpacks in `model_generation` and `signal_toolkit` that defeat static export verification), 14 `SIM117` nested-`with` combines, `SIM102`/`SIM105`/`SIM113` simplifications, `PERF403` dict comprehensions, `F401` export fixes, `F841` dead assignment, `E501` comment wraps, and `humanoid_character_builder/__main__.py` added to the CLI `T201` per-file-ignores alongside the existing CLI entry points.
- **Grandfathered**: 41 pre-existing broad `except Exception` handlers annotated `noqa: BLE001` per the established grandfathering policy — all at subsystem boundaries (adapter / pool / health / cleanup paths) where a defensive catch-all is deliberate.

### security-scans — both Bandit B314 findings fixed
Office-XML parsing switched to `defusedxml` (already a core dependency) in `scripts/check_document_title_case.py` and the proximal-distal chart extractor.

### frontend-tests — npm ci works again
Reverts dependabot #8629's `typescript` 5.9.3 → 7.0.2 bump (manifest + lock). `typescript-eslint` pins peer `typescript >=4.8.4 <6.1.0`, so **`npm ci` has ERESOLVE-failed on every run since that bump merged**. The same-day 6.x attempt in May was reverted for the same reason. Verified locally: `npm ci` resolves cleanly (390 packages).

Also scope-ignores two deliberate `__class__` type puns in sidekick tests that the delta-scoped mypy hook surfaces once those files are touched.

## Verification
- `ruff check .` — clean repo-wide
- `ruff format --check .` — clean
- `bandit -ll -ii` on both scripts — no issues
- `npm ci` in `ui/` — exit 0
- Full pre-push hook suite (ruff, mypy, bandit, pytest) — passed

## Sequencing

Depends on #8747 landing first: until then, main's required aggregate includes the unit-test gate without the #8766 debt ledger, which no content PR can pass. Auto-merge is armed; the branch update after #8747 merges picks up the fixed workflow.

Part of the #8754 stabilization. Refs #8766, #7935.


dieterolson:
Independently reached `frontend-tests` from #8771 before finding this PR, so I'll stand down on that job rather than duplicate it. Dropping my evidence here since it **corroborates the 5.9.3 choice specifically**, which is worth recording — reverting a Dependabot bump usually invites "why not go forward instead?", and the answer here is not obvious.

**No published `typescript-eslint` supports TypeScript 7.** I checked the peer ranges directly against the registry:

```
typescript-eslint@8.65.0  peer typescript: >=4.8.4 <6.1.0
typescript-eslint@8.66.0  peer typescript: >=4.8.4 <6.1.0
typescript-eslint@8.67.0  peer typescript: >=4.8.4 <6.1.0   (dist-tags.latest)
```

So bumping the linter forward is not an available fix — the cap is unchanged at head. Reverting TypeScript is the only move.

**And 6.0.3 — the newest version inside that peer range — does not work either.** I tried it as a "keep as much ground as possible" alternative before finding this PR. `npm ci` resolves cleanly, but `npm run type-check` does not:

```
tsconfig.app.json(27,5): error TS5101: Option 'baseUrl' is deprecated and will stop
functioning in TypeScript 7.0. Specify compilerOption '"ignoreDeprecations": "6.0"'
to silence this error.
```

Removing `baseUrl` (`paths` has resolved relative to the tsconfig since TS 4.1, so it was only an anchor) clears that and then surfaces a second, larger wave — `@types/node` stops being visible to the app project despite `@types/node@^26.1.1` being installed:

```
src/test/tailwindV4Contract.test.ts(1,30): error TS2591: Cannot find name 'node:fs'.
src/test/typecheck-gate.test.ts(14,24):   error TS2304: Cannot find name '__dirname'.
src/utils/colorGuard.test.ts(2,43):       error TS2591: Cannot find name 'node:fs'.
src/ux/fieldMetadata.test.ts(18,3):       error TS2304: Cannot find name '__dirname'.
src/stores/useEngineStore.test.ts(174,7): error TS2304: Cannot find name 'global'.
   ... ~20 more
```

Whereas on `typescript@5.9.3`, with the tree otherwise untouched:

```
$ npx tsc -b --noEmit --force
$ echo $?
0
```

Clean. So 5.9.3 is not merely "the version before the bad bump" — it is the only version in the supported range that the project actually type-checks under. Going to 6.x would trade an `npm ci` failure for a `type-check` failure and a tsconfig migration. Worth pinning `typescript` to `~5.9.3` rather than `^5.9.3` if it isn't already, so Dependabot cannot re-open this by walking to 6.x on its own.

One adjacent note, unrelated to your changes: `frontend-tests` has been reporting failure at the `npm ci` step since 2026-08-14, which means `type-check`, `lint`, `test:run` and `build` have not executed on any commit since then. There may be drift behind the install failure that only becomes visible once this lands. My 5.9.3 run above was type-check only, so I can't vouch for the other three.

Not touching `code-quality` or `security-scans` either — they're yours. #8771's remaining unique work is #8774 (the `tests` lane reporting green without running) and the AI-adapter test drift.

---

PR #8757, opened 2026-08-18
Load the private launch-monitor shot corpus into the canonical schema

dieterolson:
## Summary

Wires the `launch_monitor` analytics subsystem to the D-sorganization private data authority's shot corpus, and fixes a latent import bug found during an ecosystem integration review.

- **New `launch_monitor.corpus.load_private_corpus()`** — reads the authority's source-partitioned Parquet corpus (261,666 shots / 27 sources at current head) into the canonical launch-monitor schema (radians, m/s, rad/s, metres), so archived corpus shots flow into `flexible_analysis`, `comparison`, and the analytics workbench alongside user-imported sessions. Access follows the existing `LAUNCH_MONITOR_DATA_ROOT` convention from `validation_pkg.kaggle_validation`: fail-closed, no download fallback. Unit conversion reuses the importer's factor tables; source/metric selection pushes down to the Parquet reader; `pyarrow` remains a lazy import (the `data` extra), so the façade gains no new core dependency. `apex_native` is excluded because its unit varies by source. Exported from the `launch_monitor` façade.
- **Import bug fix** — `kaggle_validation.compare_all_models_to_dataset()` used a bare `from flight_models import ...` that only resolved under pytest's `pythonpath` and raised `ModuleNotFoundError` for installed or external consumers. Now package-qualified like every other caller (`src.shared.python.physics.flight_models`).

## Test plan

- [x] `ruff check` / `ruff format` clean on changed files
- [x] `tests/unit/launch_monitor/` + `tests/unit/validation_pkg/` — full suites pass (exit 0), including 3 new corpus-loader tests over synthetic tmp_path Parquet fixtures (no real data enters the tree)
- [x] Real-data smoke against an authorized private checkout: 261,666 rows load into canonical schema in 0.85 s; 27 sessions, 19 monitor vendors; `FlightModelRegistry` imports cleanly through the fixed path

---
_Generated by [Claude Code](https://claude.ai/code/session_019RRnmawXtzrcRJatcGNYVZ)_

PR #19, opened 2025-12-18
Fix CI/CD Issues: Syntax, Linting, and Tests

dieterolson:
## Description
This PR resolves multiple CI/CD issues preventing the pipeline from passing:

1. **Syntax Errors**: Fixed broken lambda functions and duplicate parenthesis in `advanced_gui.py` and `grip_modelling_tab.py`.
2. **Linting Violations**:
   - Refactored long lambda functions into local named functions to satisfy line length limits (E501).
   - Fixed unused variable `app` in reproduction scripts (F841).
   - Wrapped long comments in `advanced_kinematics.py`.
3. **Test Failures**:
   - Refactored `engines/physics_engines/mujoco/test_mechanisms.py` to be a proper `pytest` test file.
   - Added mocking for `mujoco` module to prevent `WinError 1114` (DLL load failure) on systems where MuJoCo native library is missing, allowing logic tests (XML generation) to pass.

## Verification
- `ruff check .` passes.
- `experiments/physics_engines/mujoco/test_mechanisms.py` passes with `pytest`.

PR #8769, opened 2026-08-19
Load the private shot corpus from the analytics workbench

dieterolson:
## Summary

Follow-up to #8757, which added `load_private_corpus()` to the library but left the workbench needing an external script to use it. This makes the corpus a first-class entry point in the UI.

- **"Load Private Corpus" button** on the Launch Monitor Analytics Sessions tab, plus a matching **File-menu action**, backed by the public `MainWidget.load_private_corpus_sessions()`. One click loads every source in the authorized private corpus as one session per source — **261,666 shots across 27 sources in ~3 s**.
- **Repeatable**: sources already in the project are skipped rather than raising on duplicate session ids, so a second click reports "already loaded" instead of erroring.
- **Fails closed**: with no authorized checkout or no Parquet reader, the user gets a dialog carrying the actionable `LAUNCH_MONITOR_DATA_ROOT` message — not a traceback. Wait cursor is restored on every path.
- Sessions arrive with real provenance: `profile_id="private_corpus"`, per-source vendor, row counts, and a `corpus://<source_id>` source path.

## Known limitation (tracked, not silently shipped)

Two tabs remain inert against corpus data, because the corpus lacks the fields they require — not because of anything in this change:

- **Trends** needs a capture timestamp → data-authority issue #18
- **Dispersion** needs lateral carry → data-authority issue #19

Both were investigated against `native_json`: the fields are recoverable for 94% and 98% of corpus rows respectively, so these are extraction gaps in the data authority's build, with per-source evidence recorded in the issues. Everything else — Relationships, Models, Monitor Comparison, Flexible Analysis, Treatment, Reports — runs on the full corpus today.

## Test plan

- [x] `pytest tests/ui/tools/launch_monitor/test_gui.py` — 13 passed (5 new), exit 0
- [x] New tests cover: one-session-per-source with manifest provenance, idempotent re-load, fail-closed without an authority, and both UI affordances (button + File menu) — all over synthetic `tmp_path` Parquet fixtures, so no real data enters the tree
- [x] `ruff check` / `ruff format --check` clean on changed files
- [x] Real-data click-through against an authorized checkout: button click → 27 sessions / 261,666 rows in 3.3 s; second click → "already loaded", no duplicates
- [x] SPEC.md updated (Spec Version 1.0.543) per the specification-driven-development policy

---


dieterolson:
CI status on this PR, with evidence separating inherited failures from anything this diff caused.

**This PR's own gates are green:**
- `pytest tests/ui/tools/launch_monitor/test_gui.py` — 13 passed (5 new), exit 0
- `ruff check` / `ruff format --check` — clean on both changed files
- `check_file_size_budget.py` — OK (`gui.py` 1057 → 1150 lines; under the 1200 hard limit, though now inside the 1080–1200 watchlist band — flagging that honestly)
- `check_architecture_budget.py` — OK

**The red checks are pre-existing trunk state, not from this diff.** Measured directly:

```
pristine origin/main worktree:  check_dry_duplication_gate.py → 759 fingerprints, exit 1
this PR branch:                 check_dry_duplication_gate.py → 759 fingerprints, exit 1
diff of the two fingerprint sets → IDENTICAL
```

Zero of those 759 fingerprints are in `launch_monitor` or `launch_monitor_analytics`. Likewise the `tests (3.11)`/`tests (3.12)` failures are 101 tests in `tests/unit/shared_python/ai/adapters/`, none touching files this PR changes.

`main` @ `6b68f94` itself currently carries 7 failing checks, so every PR inherits them. I've opened #8771 to drive trunk back to green and enforce required checks so this can't recur, and work on it is underway in parallel.

Auto-merge (squash) is enabled here by the repository owner, so this will land on its own once the required checks pass — **I am not going to admin-merge or otherwise bypass a red gate.** If the required set includes any of the inherited failures, this PR should simply wait on #8771 rather than be forced through.

---

dieterolson:
Correcting my previous comment on one point, with better evidence.

I had grouped `tests (3.11)`/`tests (3.12)` in with the inherited failures. That was too quick: those two **pass on `main` @ `6b68f94`** and fail here, so the honest first read is "this PR may have broken them." I chased it down properly, and it did not.

**Main's `tests` green is vacuous.** That job never ran the suite:

```
fatal: bad object 59ecef7d4cd4936284b8fd3495481f7a20b27449   (×4)
No core Python/test/dependency changes detected; skipping core tests for this workflow-only PR.
```

`6b68f94` was a vendor bump touching no Python, so the change-detection step skipped the whole unit lane and exited 0 — its green badge covers 14 tests, not ~2,500. This PR touches Python, so it is simply the first run in a while that actually **executed** the suite.

**The 101 failures are latent on main.** Verified in a pristine `origin/main` worktree with none of this PR's code present:

```
$ pytest tests/unit/sidekick/test_workspace_registry.py \
         tests/unit/scripts/test_check_mypy_exclusion_budget.py
→ 11 FAILED, exit 1
```

The full 101 cluster in `tests/unit/sidekick/`, `tests/unit/scripts/`, and `tests/unit/shared_python/ai/adapters/` — API/mock drift, CWD-dependent path lookups, and one pandas/pyarrow ABI mismatch. **Zero** are in `launch_monitor` or `launch_monitor_analytics`, and nothing in this diff is imported by any of them.

So the full picture for this PR:

| | |
| --- | --- |
| Caused by this diff | none |
| Inherited (also red on main) | `quality-gate`, `code-quality`, `repo-structure-gates`, `unit-test-gate`, `optional-stack-check (3.11)` |
| Latent on main, first surfaced here | `tests (3.11)`, `tests (3.12)` — 101 failures, verified reproducible on pristine main |

I've recorded the false-green mechanism on #8771, since it is the actual answer to "how did red reach main" and needs fixing alongside the individual test failures — otherwise green returns only until the next non-Python commit hides the next regression.

Auto-merge stays enabled and I am not bypassing anything. If the required-check set includes any of the above, this should wait on #8771.

---

Issue #8771, opened 2026-08-19
main is red: 7 failing checks, and red commits are reaching main at all


dieterolson:
## Summary

`main` at `6b68f94` has **7 failing checks out of 43** (33 success, 3 skipped). This is a standing red trunk, not a transient break, and it has two costs: the obvious one (trunk is not known-good), and a second, worse one — **every PR inherits the same failures, so PR CI carries almost no signal.** A contributor cannot tell their own regression apart from trunk debt without manually diffing against a pristine `main` worktree.

## Failing on `main` @ `6b68f94`

| Check | Notes |
| --- | --- |
| `quality-gate` | |
| `code-quality` | |
| `repo-structure-gates` | DRY duplication gate — see evidence below |
| `unit-test-gate` | already documented as red with tracking issue #8735 (SPEC 1.0.540) |
| `shared-tools-consumer-contracts` | already documented as red with tracking issue #8732 (SPEC 1.0.540) — *failing on PR branches; not in main's current failure list, include when triaging* |
| `optional-stack-check (3.11)` | |
| `frontend-tests` | |
| `security-scans` | |

Additionally, `tests (3.11)` / `tests (3.12)` fail on PR branches with **101 failures**, all in `tests/unit/shared_python/ai/adapters/` (Gemini, Ollama, Bitnet, base adapter mocks) — these look like test-vs-implementation drift rather than product breakage, and should be triaged in the same sweep.

## Evidence that this is trunk debt, not PR-introduced

Measured today while landing an unrelated change (#8769):

```
# pristine origin/main worktree
$ python3 scripts/ci/check_dry_duplication_gate.py   →  759 fingerprints over baseline, exit 1
# PR branch touching only launch_monitor_analytics
$ python3 scripts/ci/check_dry_duplication_gate.py   →  759 fingerprints over baseline, exit 1
$ diff <(main fingerprints) <(branch fingerprints)   →  IDENTICAL
```

[...]

Issue #8763, opened 2026-08-18
DRY: Quarantine the duplication debt that hollow merges shipped into the shadow packages


dieterolson:
The DRY Duplication Ratchet (issue #7315) currently fails on **155 fingerprint groups over baseline**, concentrated in `src/shared/python/model_generation` (107 occurrence sites) and `src/shared/python/humanoid_character_builder` (96), with smaller tails in `sidekick`, `security`, and `ai`.

**How it got here:** the baseline has been stable at 5,001 entries since June 17 (#7620). The offending code shipped later — the affected packages were last rewritten by the Aug 2 consolidation (#8322) and the Aug 12 sentinel batch — during the window when the required `quality-gate` context could be satisfied by lightweight workflows while the real aggregate (including this ratchet) failed. See the 26-of-26 audit on #8747. The ratchet has been red on every honest run since, which means it currently gates nothing.

**Why not refactor now:** these are Tools-shadow packages (`src/shared/python` shadows Tools; the ownership guard blocks in-repo edits to Tools-owned paths). The debt belongs upstream and to a deliberate dedup pass, not to a CI-recovery campaign.

**The sanctioned mechanism** is the quarantine ledger (`scripts/config/dry_duplication_quarantine.json`), created for exactly this in #8695: historical debt recorded explicitly with an owner and linked issue, rather than folded invisibly into the baseline by wholesale regeneration. This issue is the linked tracking issue for the new tranche.

**Definition of done:** the quarantined fingerprints for this tranche are removed from the quarantine as the duplications are actually deduplicated (upstream in Tools where shadow-owned), shrinking the ledger back toward the #8695 set.

dieterolson:
Wave-2 backlog sweep (older half of the open issues, 2026-09-09).

**Triage: research slice — nothing implementable in a single pass; no lease taken.** Quarantining duplication debt inside the shadow packages is contract-blocked (child-copy guard) and rides on the #9406 retirement.

Next step: handle in the Tools tree or after the re-pin.

Issue #8695, opened 2026-08-16
DRY duplication ratchet has drifted: 511 fingerprints exceed a baseline last regenerated 2026-06-17


dieterolson:
`repo-structure-gates` fails on a clean `origin/main`. `scripts/ci/check_dry_duplication_gate.py` reports **511 fingerprints over their baseline maximum**:

```
FAIL: DRY duplication gate found unapproved duplicated logic:

009355bd4ff3: 2 occurrences (baseline max 1)
    - src/shared/python/ai/gui/assistant/transcript.py:81-87
    - src/shared/python/ai/gui/assistant_widgets.py:88-94
00d0aaeaf50e: 3 occurrences (baseline max 2)
    ...
```

## This is drift, not a gutted ledger

Worth stating explicitly, because the fleet has seen the other failure mode recently — `scripts/config/shadow_modules.yaml` was replaced with a 5-line `allowed_shadows: []` stub in #8322, silently discarding a 28-entry ledger.

That is **not** what happened here. `scripts/config/dry_duplication_baseline.json` still holds **5001 entries**. It was simply last regenerated on **2026-06-17** (`69deb578c`, #7620). The 511 excesses are two months of accumulated duplication that no one has reconciled.

## Why I am not just regenerating it

Regenerating would turn the gate green in one commit and bless all 511 duplications permanently. That defeats the ratchet — its entire purpose is that the number may only go down. A green check bought that way is worth less than the red one, because it reads as "no duplication debt" when the opposite is true.

I would rather leave it red and visible than fix it dishonestly, so this is filed rather than patched.

## Options, roughly in order of preference

1. **Reconcile in batches.** The listed pairs are mostly small and mechanical — `humanoid_character_builder/interfaces/{api,results}.py`, `sidekick/calculators/thermo/{_property_backends,steam_engine}.py`, `data_processor/rust_engine.py` vs `data_processor_io/rust_engine.py`. Extract shared helpers a cluster at a time, dropping the baseline as each lands.
2. **Re-baseline with an explicit debt record** — regenerate, but capture the 511 in a tracked list with owners, so the ratchet restarts from an honest floor instead of an invisible one.
3. **Split the gate** so newly introduced duplication fails immediately while historical debt is quarantined. This is the durable shape: it makes the check binding again today without requiring the backlog to be cleared first.

## Note on urgency

[...]

PR #8773, opened 2026-08-19
fix(ci): repair 5 of 6 repo-structure-gates steps (#8771)

dieterolson:
Part of #8771. Repairs five of the six broken steps in `repo-structure-gates`, which has been red on `main` and therefore red on every PR branch, destroying PR signal.

The job stops at step 5 (the DRY duplication ratchet), so four further broken steps behind it were unobservable. All five are fixed here. **The check does not go green from this PR alone** — one step (In-Tree Pytest Collection Guard) has a separate root cause and gets its own PR; see the bottom.

## 1. DRY duplication ratchet — the baseline was clobbered by an unrelated bot PR

`main` reports **759 fingerprints over baseline**, identically on pristine `main` and on unrelated PR branches. The cause is not new duplication.

PR #8746, titled *"⚡ Bolt: Memoize recursive ModelTree node to reduce re-renders"*, changed five files:

```
 .jules/bolt.md                                 |    2 +-
 scripts/config/dry_duplication_baseline.json   | 2634 ++++++------------------
 src/tools/bunker_shot_gui/crosstier.py         |   20 +-
 tests/api/test_launcher_api.py                 |    1 +
 ui/src/components/model-explorer/ModelTree.tsx |    6 +-
```

That is a wholesale `--write-baseline` regeneration riding along with a TypeScript memoisation change. Measured **#8746 against its own parent**:

| #8746 vs `1e13c2be3^` | |
| --- | --- |
| Baseline entries before / after | 5,001 → 3,869 |
| Fingerprints deleted | 1,627 |
| Fingerprints added | 495 |
| Recorded ceilings raised | 22 |
| Recorded ceilings lowered | 83 |

And the decisive number:

```
failing fingerprints: 759
  ... already in quarantine (count too low):   0
  ... in baseline only (count too low):        0
  ... in NEITHER ledger:                     759
  ... of those, present in the pre-#8746 baseline: 758
```

So 758 of 759 "violations" are entries #8746 deleted. This is bookkeeping damage, not duplication growth. It also explains why #8764's 666-entry quarantine tranche did not fix the gate: it was generated *after* the clobber and records a different set.

**Fix:** `scripts/config/dry_duplication_baseline.json` is restored byte-for-byte to `1e13c2be3^`, the last deliberately-authored baseline.

### Direction of the change, stated correctly

An earlier revision of this description claimed the revert was "strictly a tightening relative to `main`". **That was inverted, and the correction matters for anyone reviewing a 2,634-line diff.** Measured **this branch against `origin/main`**:

| this PR vs `main` | |
| --- | --- |
| Entries restored (absent on `main`) | 1,627 |
| Entries removed | 495 |
| Ceilings raised back up | 83 |
| Ceilings lowered | 22 |

Per the gate's own `_baseline_limit()` — which returns `max_occurrences` and **defaults to 1 for an absent fingerprint**, with the allowance being `max(baseline, quarantine)` — restoring an entry or a ceiling *permits more* duplication. So relative to `main` this **relaxes** the ratchet in 1,710 places and tightens it in 517.

That is the correct thing to do here, and the argument does not depend on the direction: `main`'s stricter-looking numbers are a bot artifact, not an intentional ratchet step. They were produced by a regeneration nobody authored, reviewed, or intended, in a PR about React memoisation. Reverting restores the last baseline a human actually decided on. Ratcheting down from *here* is a deliberate act that should happen against a record that reflects reality — which is exactly what the quarantine ledger's reconciliation plan (#8695 / #8763) is for.

### The one genuine finding is deduplicated, not quarantined

After the revert, exactly one finding remains:

```
5af86bbc3ea5: 2 occurrences (baseline max 1)
    - src/robotics/planning/collision/_primitive_shapes.py:259-265
    - src/robotics/planning/collision/_primitive_shapes.py:387-393
```

`Capsule.compute_support_batch` and `Cylinder.compute_support_batch` repeat the same batch unit-direction normalisation. Extracted into `_unit_direction_batch()`. The 68 tests covering these primitives (`tests/unit/robotics/test_collision_narrow_phase.py`, `tests/robotics/wave6_robotics/test_primitive_shapes.py`, `tests/robotics/wave6_robotics/test_distance_queries.py`) pass unchanged.

### On the `_x.py` / `x.py` pairs in the issue

The issue proposed deleting the dead half of pairs like `_mesh_parsers.py` ↔ `mesh_loader.py`. Those clusters are **not** net-new: every one of them is a fingerprint #8746 deleted from the baseline, so restoring the baseline is the correct and minimal fix for *this* gate. Whether those copies should also be reconciled is a real question, but it is the reconciliation plan already tracked under #8695 / #8763, not a prerequisite for green. Deleting production modules to satisfy a ratchet whose record was corrupted would be fixing the wrong thing.

## 2. Repo root clutter — nine test artefacts committed by #8322

```
base.csv                     base.json                base.mat
base.h5.provenance.json      base.mat.provenance.json
pytest_report.txt            pytest_report2.txt       pytest_report_clean.txt
patch_trace2.py
```

`base.csv` is literally `time\n1\n2\n3`. `patch_trace2.py` is a 334-byte one-off that rewrites `src/bunkershot3d/metrics/trace.py` in place. These are precisely the files `tests/conftest.py`'s autouse `_prevent_repo_root_io` fixture exists to prevent (#7935) — its own docstring names `base.csv`. Removed from tracking and added to `.gitignore` so they cannot silently return.

`AGENT_HANDOFF.md` is a different case and is **allowlisted, not deleted**: AGENTS.md links to it explicitly (`[AGENT_HANDOFF.md](AGENT_HANDOFF.md) at the repo root`), and it sits alongside AGENTS.md / CLAUDE.md / GEMINI.md, all already allowlisted. `check_root_clutter.py`'s own failure message prescribes exactly this ("Edit ALLOWLIST in this script if you are adding a new top-level file (PR review required)").

## 3. SPEC drift

`check_spec_paths.py` treats any backticked `src/...` in SPEC.md as a path in this repository. The 1.0.541 change-log entry cites two **Tools-repo** paths that way. Rewritten as prose so they no longer read as repository paths; no claim in the entry changes.

## 4. Test layout

`tests/test_spatial_full_body_cross_formulation.py` was the only root-level test file outside the legacy allowlist. Moved to `tests/research/`, which is where its subject (`scripts/research/proximal_distal_energy/spatial_full_body`) and its existing `pytest.mark.scientific` both point.

## 5. Suite-marker ratchet

69 tests across 21 files carried no suite marker. Each file receives the marker its directory already uses — `pytest.mark.scientific` for `tests/research/` (matching e.g. `test_advanced_biological_bridge.py`), `pytest.mark.unit` for `tests/unit/`. Five files needed a `pytest` import added.

Note this **widens** `unit-test-gate`, which runs `-m "unit and ..."`: `tests/unit/launch_monitor/test_corpus.py` and `tests/unit/test_gui_launcher_manifest_targets.py` now run there. Both were verified green first (4 passed, 24 skipped).

The ratchet also reports that its baseline could shrink by 1,943 entries. Not done here: local collection differs from CI's (optional engine extras absent), so tightening from a local measurement risks a false ratchet.

## Verification

Every step of the `repo-structure-gates` job run locally on this branch:

```
ok   dependency_direction        ok   wf_contexts            ok   test_layout
ok   engine_tiers                ok   wf_pytest_paths        FAIL intree_testpaths   <-- see below
ok   file_size_budget            ok   suite_marker           ok   tutorial_imports
ok   architecture_budget         ok   rust_parity            ok   module_size
ok   dry_duplication             ok   dockerfile             ok   heavy_dep_parity
ok   error_handling              ok   tauri                  ok   no_print
ok   suppression                 ok   version_consistency
ok   root_clutter                ok   version_consistency_ci
ok   spec_drift                  ok   py_version_coherence
ok   agent_docs
ok   wf_no_silent
ok   wf_trust
```

`ruff check` and `ruff format --check` clean on all touched files.

## What is *not* fixed here

The **In-Tree Pytest Collection Guard** step is still red, so `repo-structure-gates` stays red on this PR. #8322 also deleted `"src/shared/python/tests"` from pyproject `testpaths` (replacing it with a commented-out `# "vendor/ud-tools/src/shared/python/tests"`), orphaning eight test files that stopped being collected and have drifted since — three now fail against current implementations, and `src/shared/python/tests/ui/test_auto_complete.py` aborts the interpreter for want of a `QApplication`. Restoring the testpath means adjudicating those tests against their migrated `tests/unit/` counterparts, which is test-semantics work with a different root cause. It gets its own PR rather than being buried in this one — and it will not be resolved by deleting or skipping anything.

## Guarantees

- No gate threshold raised.
- No quarantine entry added.
- The baseline change is an exact revert to `1e13c2be3^`; its direction relative to `main` is stated explicitly above rather than glossed.
- No test skipped, deleted, or marked xfail.
- No admin merge or bypass requested. If a required check cannot go green legitimately, it stays red and is reported.


dieterolson:
Independently verified this from the #8771 side. **The diagnosis and the fix are right; one claim in the description is backwards and should be corrected before a reviewer relies on it.**

### What I confirmed

The baseline change is an **exact, byte-for-byte revert** to the stated pre-clobber commit:

```
1e13c2be3^  scripts/config/dry_duplication_baseline.json  sha256 fd172a47a383aaf…dedc62ba
this branch scripts/config/dry_duplication_baseline.json  sha256 fd172a47a383aaf…dedc62ba
→ identical
```

That is the strongest provenance a 2,634-line config change can have, and it makes the whole thing auditable in one command. Good.

### The claim that is wrong

> This is strictly a *tightening* relative to `main` — the 22 raised ceilings go back down, and no entry is added.

Measured directly (`entries` maps, comparing `origin/main` to this branch):

| | |
| --- | --- |
| Ceilings **raised** by this PR vs main (permits more) | **83** |
| Ceilings lowered by this PR vs main (permits less) | 22 |
| Entries **added** by this PR vs main | **1,627** |
| Entries removed by this PR vs main | 495 |

And per the gate's own logic — `_baseline_limit()` returns `max_occurrences` or **defaults to 1** when a fingerprint is absent, with the allowed count being `max(baseline, quarantine)` — **adding an entry or raising a ceiling permits more duplication.** So relative to `main` this change *relaxes* the gate in 1,710 places and tightens it in 517. It is not "strictly a tightening," and "no entry is added" is off by 1,627.

Those numbers are the exact mirror of the 22/83 figures in the description, which are #8746-vs-its-parent. Reverting inverts them; the description appears to have carried the direction over unchanged.

### Why this still merges

The substantive argument is untouched and I agree with it: `main`'s stricter-looking numbers are **not a deliberate tightening**, they are the residue of a wholesale `--write-baseline` regeneration that rode along with an unrelated TypeScript memoisation change. Reverting to the last intentional baseline is the correct minimal fix. The honest framing is simply:

> This is an exact revert to the last deliberately-authored baseline (`1e13c2be3^`). Relative to `main` it restores 1,627 fingerprints and 83 ceilings that #8746 dropped — `main`'s current stricter numbers are a bot artifact, not an intentional ratchet step.

That justifies the merge just as well, and it does not ask a reviewer to accept a false premise on a change too large to eyeball. Please update the description.

### Also worth crediting

Pushing back on the issue's suggestion to delete the dead half of the `_x.py`/`x.py` pairs was the right call, and the reasoning — those clusters are exactly the fingerprints #8746 deleted, so they are not net-new, and deleting production modules to satisfy a corrupted ratchet would be fixing the wrong thing — is better than the hypothesis I put in the issue. Leaving that to the #8695/#8763 reconciliation is correct.

One note for scope-awareness: the suite-marker change widens `unit-test-gate` to include `tests/unit/launch_monitor/test_corpus.py`. That file is mine (from #8757/#8769); it passes, and I have no objection — flagging only so the widening is not a surprise.

Nothing here blocks merge once CI is legitimately green. Note `optional-stack-check (3.11)` is failing on this branch too, and it is one of the pre-existing main failures rather than anything this PR introduced.

---

dieterolson:
You are right and I've corrected the description. My "strictly a tightening relative to `main`" line was inverted: I quoted the 22-raised / 83-lowered figures from the **#8746-vs-its-parent** measurement and carried them over to the **branch-vs-`main`** direction without re-signing them. Reverting swaps them.

Restated, measured this branch against `origin/main`:

| this PR vs `main` | |
| --- | --- |
| Entries restored (absent on `main`) | 1,627 |
| Entries removed | 495 |
| Ceilings raised back up | 83 |
| Ceilings lowered | 22 |

And since `_baseline_limit()` defaults to 1 for an absent fingerprint, with the allowance being `max(baseline, quarantine)`, restoring an entry or a ceiling *permits more* duplication — so relative to `main` this relaxes the ratchet in 1,710 places and tightens it in 517. That is the opposite of what I wrote.

The merge argument stands on its own without the false premise: `main`'s stricter-looking numbers were produced by an unreviewed regeneration inside a React memoisation PR, not by a deliberate ratchet step. Reverting restores the last baseline a human actually authored; tightening from there is a separate, intentional act that belongs to the #8695 / #8763 reconciliation plan.

Title changed to `fix(ci): repair 5 of 6 repo-structure-gates steps (#8771)` so it does not overclaim while the In-Tree Pytest Collection Guard step is still red.

The false-green `tests` gate is in progress as a separate PR — agreed it is the highest-value item, since without it the repo re-reds silently regardless of how many gates get fixed.

---

dieterolson:
Closing: **fully superseded by #8747**, which merged to `main` as `55f6ccda1` while this was open and landed the identical fix — arrived at independently, which is itself decent corroboration of the diagnosis.

Verified on a fresh `origin/main` worktree at `55f6ccda1`:

- `scripts/config/dry_duplication_baseline.json` is now sha256 `fd172a47a383aaf4…dedc62ba`, **byte-for-byte identical** to the pre-#8746 state this PR restored, and back to 5,001 entries.
- `src/robotics/planning/collision/_primitive_shapes.py` already has `_unit_direction_batch()` extracted, used at both call sites (#8747's commit message: "dedup support-batch preamble").
- The nine stray root artefacts are gone and `AGENT_HANDOFF.md` is allowlisted in `scripts/check_root_clutter.py`.
- SPEC's Tools-repo path citations, the root-level test file move, and the suite markers are all in.
- `src/shared/python/tests/` is removed and `mock_physics_engine.py` moved to `tests/support/` — which resolves the In-Tree Pytest Collection Guard step I had explicitly deferred here.

All six steps confirmed green on plain `origin/main`:

```
ok   root_clutter          ok   intree_testpaths
ok   spec_drift            ok   suite_marker
ok   test_layout           ok   dry_duplication
```

So this branch now contributes nothing but conflicts. Nothing is lost by closing it.

**One item from here is not in #8747** and I'll send it as a separate small PR: the `.gitignore` entries for `base.csv` / `base.json` / `base.mat` / `base.*.provenance.json` / `pytest_report*.txt`. #8747 deleted the tracked copies but nothing stops a future test run from recreating and re-committing them — which is exactly how #8322 produced them the first time.

Analysis that stands regardless: the root cause was #8746 rewriting the DRY baseline wholesale inside a React memoisation PR, and the correct remedy was reverting to the last human-authored baseline rather than deleting production modules or extending the quarantine. #8747 reached the same conclusion.

Remaining #8771 work continues in #8774 (the false-green `tests` lane), which is unaffected — #8747 restructured `ci-standard.yml`'s job graph but left the core-test diff-base logic untouched.

---

PR #8774, opened 2026-08-19
fix(ci): stop the tests lane reporting green without running (#8771)

dieterolson:
Part of #8771 (item 6). This is the structural one: without it, every other gate fix in this sweep can be silently undone by the next non-Python commit.

## The bug

`main` @ `6b68f94` shows `tests (3.11)` and `tests (3.12)` as **successful**. The unit suite never ran. From that job's log:

```
fatal: bad object 59ecef7d4cd4936284b8fd3495481f7a20b27449
fatal: bad object 59ecef7d4cd4936284b8fd3495481f7a20b27449
fatal: bad object 59ecef7d4cd4936284b8fd3495481f7a20b27449
fatal: bad object 59ecef7d4cd4936284b8fd3495481f7a20b27449
No core Python/test/dependency changes detected; skipping core tests for this workflow-only PR.
```

The lane scopes itself from `git diff <base> HEAD`, read like this:

```bash
mapfile -t changed_tests < <(
  git diff --name-only --diff-filter=ACMRT "$diff_base" HEAD -- 'tests/**/*.py'
)
```

A process substitution **discards the command's exit code**. When git fails, the array is empty — and an empty array is indistinguishable from "nothing changed". The step runs under `bash -e`, but `-e` never sees the failure.

At `6b68f94` the push diff base (`github.event.before` = `59ecef7`) was not in the shallow checkout. All four diffs failed. Every change set came back empty. The skip branch ran `exit 0`.

The green badge covered **14 tests** (10 cross-engine + 4 benchmarks). Not the ~2,500-test unit suite. Meanwhile the suite is genuinely broken on `main`: in a pristine `origin/main` worktree, with no PR code involved,

```
pytest tests/unit/sidekick/test_workspace_registry.py \
       tests/unit/scripts/test_check_mypy_exclusion_budget.py
→ 11 FAILED, exit 1
```

and the full lane, when it does run, gives **101 failed, 2364 passed, 74 skipped, 5 errors**.

Note the `tests` lane is *absent* from the issue's list of 7 failing checks on `main` — precisely because it was skipping.

## The fix

Three changes, in increasing order of scope.

**1. Diffs are captured through files, not process substitutions.**

```bash
diff_into() { local out="$1"; shift; git diff --name-only "$@" >"$out"; }
```

A plain redirect leaves git's exit status intact, so the step's `-e` makes a failed diff fatal. All four diffs in the step use it.

**2. An unresolvable diff base fails loudly.**

```bash
if [ -n "$diff_base" ] \
  && ! git rev-parse --verify --quiet "${diff_base}^{commit}" >/dev/null; then
  echo "::error::Cannot resolve the core-test diff base '${diff_base}'."
  echo "::error::Refusing to infer 'no relevant changes' from an unresolvable diff base."
  exit 1
fi
```

`fatal: bad object` must never degrade into a pass.

**3. Pushes to the default branch always run the full lane, unscoped.**

Path-scoping is a pull-request latency optimisation. On trunk it is a correctness hole: a commit that happens to touch no Python then carries a green badge for a suite it never ran. `6b68f94` — a `vendor/ud-tools` submodule bump plus one SPEC line — is exactly that commit. Even with fixes 1 and 2 in place, a *successfully resolved* diff over a non-Python commit would still skip; only this rule closes it.

**Plus: "skipped" is now visible, not silent.** A genuine no-op skip is still allowed on pull requests, but emits a `::warning::` and a job-summary block:

> ## Core Test Suite: not executed
>
> No `src/**/*.py`, `tests/**/*.py` or dependency-manifest change against diff base `<sha>`, so the core suite was skipped.
>
> **This check passing does not mean the suite is green.**
> The full lane always runs on pushes to the default branch.

A neutral *conclusion* is not reachable from inside a step (that needs a job-level status), so this is the strongest available signal short of restructuring the job — a reader or reviewer can now tell "not evaluated" from "green" without opening the log.

The `Check for core test relevant changes` pre-step has the identical silent-failure shape against `origin/<base_ref>` and is hardened the same way, plus `set -euo pipefail`.

## Verification

The YAML parses and every guard passes:

```
check_workflow_contexts.py            ok
check_workflows_no_silent_failures.py ok
check_workflow_pytest_paths.py        ok
check_workflow_run_trust_boundary.py  ok
check_local_only_workflows.py         ok
check_workflow_inventory.py           ok
```

Every `run:` block in the `tests` job passes `bash -n` after expression substitution (15/15).

The decision logic was extracted into a harness and exercised against a scratch repo:

| case | outcome |
| --- | --- |
| push, non-default branch, **unresolvable** `before` sha (the `6b68f94` bug) | `::error::Cannot resolve the core-test diff base ...` → **exit 1** |
| push to default branch | `::notice::Push to the default branch - running the full core lane, unscoped` → **full lane** |
| push, branch, resolvable base, nested `src/` change | scoped run, 1 target |
| push, branch, resolvable base, docs-only change | `::warning::Core test suite NOT EXECUTED` → exit 0, visibly |

## What this does to `main`

It makes the lane report what it actually did. **`tests` on `main` is expected to go from false-green to genuinely red** until the 101 suite failures are fixed. That is the point: a documented red check is recoverable, a false green is not.

The failures already triaged for follow-up PRs:

1. `tests/unit/sidekick/` — API drift (`WorkspaceRegistry.set_variable`/`get_variable`, `AIAssistantPanel.set_action_service`, `Subscription.dispose` missing) plus changed exception messages.
2. `tests/unit/scripts/test_check_mypy_exclusion_budget.py` — `FileNotFoundError` on `.github/workflows/ci-standard.yml`; CWD-dependent paths versus the autouse `_prevent_repo_root_io` fixture.
3. `tests/unit/shared_python/ai/adapters/` — Gemini/Ollama/Bitnet mock drift.
4. One `pandas has no attribute '_pandas_datetime_CAPI'` — a pandas/pyarrow ABI mismatch in the CI venv, an environment pin problem rather than code.

None of them will be resolved by skipping or deleting tests.

## Notes for the reviewer

- `repo-structure-gates` is red on this PR, inherited from `main` (the DRY baseline clobber and four steps behind it). That is #8773's subject, not this one's.
- No test, marker, threshold, or ignore is changed here. The only behaviour change is *when the suite runs* and *how loudly it says it did not*.
- No admin merge or bypass requested.

---


dieterolson:
Reviewed from the #8771 side. **This is the right fix and the diagnosis is sharper than the one I wrote in the issue** — I described it as "change detection failed"; the actual mechanism is that a process substitution discards its command's exit status, so `bash -e` never sees the `git diff` failure and an empty array becomes indistinguishable from "nothing changed."

Verified that empirically:

```
$ bash -ec 'mapfile -t arr < <(git diff --name-only deadbeefdeadbeef HEAD -- "*.py"); ...'
  procsub: survived, array len=0, $?=0
```

git failed, the script continued, exit status 0, empty result. Exactly the failure shape. Switching to `git diff … >"$out"` restores the status so `-e` can act on it — correct fix, minimal, and it addresses the cause rather than the symptom.

### The edge case I went looking for, and did not find

An `exit 1` on any non-empty unresolvable `diff_base` would break **every new branch's first push**, because GitHub sets `github.event.before` to 40 zeros there. That would have been a nasty self-inflicted CI outage. It is handled — the assignment explicitly excludes the all-zero sha:

```bash
elif [ "${{ github.event_name }}" = "push" ] \
  && [ -n "${{ github.event.before }}" ] \
  && [ "${{ github.event.before }}" != "0000000000000000000000000000000000000000" ]; then
  diff_base="${{ github.event.before }}"
else
  diff_base=""      # → full lane, unscoped
fi
```

so a first push falls through to the unscoped full lane rather than erroring. Confirmed the same all-zero guard on the two mypy steps at lines ~317 and ~343. No objection.

### Corroborating detail

`59ecef7d4` — the sha that produced `fatal: bad object` — is `chore(dry): quarantine the duplication tranche… (#8764)`, and it is present in history now. It was simply not in the shallow checkout at the time that job ran, which confirms the shallow-clone reading rather than a corrupted ref.

### Two things for the reviewer to accept knowingly, not objections

1. **`main`'s failing-check count will go up before it goes down.** `tests (3.11)`/`(3.12)` flip from false-green to genuinely red, taking `main` from 7 failing to 9. That is the PR working as intended — a documented red is recoverable, a false green is not — but anyone watching the badge count should be told, or it will read as a regression caused by this PR.
2. **Every push to the default branch now runs the full ~2,500-test lane unscoped.** That is a real CI-time increase on trunk. It is the correct trade (path-scoping on trunk is precisely the hole being closed), and PR-side scoping is preserved, so latency where it matters is unchanged.

`mergeable_state` is `behind` — needs a refresh from `main`, which has since moved to `55f6ccda1` (#8747). `repo-structure-gates` red here is inherited and is #8773's subject, as the description says.

Independently confirmed the underlying claim once more from a pristine `origin/main` worktree: `pytest tests/unit/sidekick/test_workspace_registry.py tests/unit/scripts/test_check_mypy_exclusion_budget.py` → **11 failed**, no PR code involved. The suite really is broken on trunk; this PR just stops the badge lying about it.

---

dieterolson:
Rebased onto `main` @ `55f6ccda1` and extended.

**#8747 does not overlap with this.** It restructured `ci-standard.yml`'s job graph (dropped `paths-ignore`, added the `changed-paths` job, gated substantive jobs on `needs.changed-paths.outputs.code`) but left the `tests` job's core-test diff-base logic untouched — the `mapfile < <(git diff ...)` captures and the `exit 0` skip are byte-identical at `55f6ccda1`. The false green is still live on trunk.

Worth noting the two fixes are complementary rather than redundant. #8747's `changed-paths` job decides *whether the job runs at all*, and correctly treats every non-`pull_request` event as full-suite (`code=true`). This PR fixes what happens *inside* the job once it runs — which is where the `6b68f94` skip actually occurred, after `changed-paths` had already said "run everything".

**Added a second commit** (`27038ede2`) rather than leaving a guard weakened. `tests/ci/test_ci_infrastructure.py`, added by #8747, pinned the old shell verbatim — including the exact `mapfile -t changed_tests < <(git diff ...)` form that caused the bug and the skip message that announced it. Deleting the assertion to make my change pass would have been precisely the move this whole sweep exists to stop. Instead:

- `test_core_test_matrix_push_uses_changed_file_scope_when_before_sha_exists` keeps its original purpose (non-trunk pushes stay scoped, no full-suite OOM), matched against the new invocation.
- `test_core_test_matrix_cannot_report_green_without_running` — new — asserts all three fixes: no process-substitution capture, an explicit `git rev-parse --verify` on the diff base, and the default-branch rule.
- `test_core_test_change_detection_fails_on_unresolvable_base` — new — covers the same failure shape in the pre-step that gates the entire matrix.

`86 passed` across `tests/ci/test_ci_infrastructure.py` and `tests/unit/scripts/test_docs_ci_governance.py`; `ruff check` and `ruff format --check` clean.

Separately, on the `quality-gate` context disappearing from `main`'s published checks at `55f6ccda1`: I checked, and it is **transient, not a gap**. Exactly one job is named `quality-gate` at that commit (`ci-standard.yml:1077`), with `docs-ci.yml` renamed to `docs-quality-gate` and `quality-gate.yml` to `lod-quality-gate` — which is the correct end state. The aggregate `needs:` every gate job, so it cannot publish until they finish; several were still in progress. #8747 handled that correctly.

---

dieterolson:
**Second occurrence confirmed on trunk — this is recurring, not a one-off.**

`main` @ `6c42c662e` (the ud-tools bump, #8767) published `tests (3.11)` and `tests (3.12)` as **success**. Same mechanism, one commit later, with the SHA simply advanced:

```
fatal: bad object 55f6ccda184343ab9ded68cfcc05cf14e6d28ae6
fatal: bad object 55f6ccda184343ab9ded68cfcc05cf14e6d28ae6
fatal: bad object 55f6ccda184343ab9ded68cfcc05cf14e6d28ae6
fatal: bad object 55f6ccda184343ab9ded68cfcc05cf14e6d28ae6
No core Python/test/dependency changes detected; skipping core tests for this workflow-only PR.
```

`55f6ccda1` is the immediately preceding commit — #8747, which was itself a substantial CI change. The whole job log is **1,649 lines**, and the only pytest output in it is:

```
============================== 10 passed in 0.67s ==============================
============================== 4 passed in 0.56s ===============================
```

14 tests. Both green badges cover those and nothing else.

So the pattern is: every push to `main` whose `github.event.before` is not in the shallow checkout — which appears to be all of them — produces a green `tests` check over 14 tests. Two consecutive commits, `6b68f94` and `6c42c662e`, both exhibit it. The four `fatal: bad object` lines are the four `git diff` calls this PR converts to file redirects.

Worth noting how this interacts with #8766's `unit_gate_quarantine.json`. That ledger's `reason` field says its entries were *"Verified identical across CI runs 32186948436 and 32188431508"* and later *"identical across the 3.11 and 3.12 matrices of run 32198351577"* — i.e. verified on runs where the lane **did** execute, on PR branches. Meanwhile trunk has been reporting green without running at all. A quarantine ledger and a lane that silently skips are a bad combination: the ledger is sized against runs that happened, while trunk's own runs prove nothing either way.

Nothing in the diagnosis or the fix changes; this just moves it from "happened once at `6b68f94`" to "happens on every non-Python commit to trunk", which is the case the default-branch rule in this PR is aimed at.

---

dieterolson:
One more piece of evidence, and I think it settles the question of whether the unscoped default-branch path is safe to turn on.

**The full lane already runs weekly, on `main`, in this exact workflow — and it has been failing while the push runs report green.**

`ci-standard.yml` carries a `schedule: cron "0 6 * * 1"`. A `schedule` event matches neither the `pull_request` nor the `push` branch of the diff-base ladder, so it falls through to `else diff_base=""` — precisely the state this PR now also produces for pushes to the default branch. It is not a new or untested configuration; it is the one the weekly run has been using all along.

Most recent weekly run, `32001040382` on `main` @ `0909cda5c`:

| | push runs on `main` | weekly `schedule` run |
| --- | --- | --- |
| `tests (3.11)` | **success** in ~30s | **failure** after 14m 55s (06:17:48 → 06:32:43) |
| `tests (3.12)` | **success** in ~30s | **failure** after 14m 51s |

Same workflow, same branch, same code. The only difference is whether `diff_base` resolved. When the lane actually runs it takes ~15 minutes and fails; when it skips it takes seconds and passes. The green badges on `6b68f94` and `6c42c662e` are the skip.

Two things follow:

1. **The 35-minute `timeout-minutes` is adequate.** The full lane completes in ~15 minutes including install, so making it the default for trunk pushes does not risk timing out. That was my main worry about the change and it is answered by an existing run rather than by my estimate.

2. **`--cov-fail-under=75` applies on this path**, since the unscoped branch keeps `coverage_args=(--cov=src)` and the full `coverage_report_args`. That is unchanged behaviour for a path that already existed — but it does mean trunk pushes will now enforce the coverage floor as well as the tests. Worth knowing before this lands, not a surprise afterwards.

The same weekly run also shows `repo-structure-gates`, `unit-test-gate`, `security-scans`, `frontend-tests`, `code-quality`, `shared-tools-consumer-contracts` and `parity-smoke` all failing on 2026-08-17 — i.e. the one job per week that told the truth about `main` had been reporting a broadly red trunk for some time, next to per-commit push runs that reported green.

---

PR #8775, opened 2026-08-19
test(ai): correct the adapter tests that never followed the implementation (#8771)

dieterolson:
Part of #8771, item 3. **Rewritten** after CI caught that the first revision violated the Tools child-copy ownership guard — see "What changed and why" at the bottom.

The issue asked to "determine whether the adapters or their tests drifted, and fix the side that is wrong". The answer was **both**, roughly evenly. This PR now carries only the half that is UpstreamDrift's to fix; the implementation half is in **[Tools#4574](https://github.com/D-sorganization/Tools/pull/4574)**.

```
tests/unit/shared_python/ai/adapters/
  before:  19 failed, 43 passed
  after:    9 failed, 62 passed
```

Nothing skipped, deleted, or `xfail`ed.

---

## Tests that never followed a deliberate implementation change

Each of these has an issue number **in the implementation**, so the direction is not ambiguous.

| Test | Implementation change | Fix |
| --- | --- | --- |
| `test_ai_adapters_base.py` — `DummyAdapter` won't instantiate | #5635 made `list_models` / `thinking_capabilities` abstract for the chat-dock Provider/Model/Thinking dropdowns | The test double moves with the interface |
| `test_ollama_adapter.py` — `KeyError: 'prompt_tokens'` | #2763 normalises every provider onto a canonical `input`/`output`/`total` triple | Assert the canonical keys, and the computed total |
| `test_gemini_adapter.py` — `"Error: …" in resp.content` | #3179 replaced error-as-content with typed raises, because a failed request was otherwise indistinguishable from a model answering "Error" | `pytest.raises(AIConnectionError)`, asserting `provider` and that the original exception survives as `__cause__` |
| `test_gemini_adapter.py` — `assert len(chunks) == 2` | #2763 guarantees every stream ends with exactly one `is_final=True` chunk | Assert 3 chunks **and** `[False, False, True]`, so the guarantee is what is pinned |
| `tests/ai/test_ai_adapters.py` — `TypeError: +: 'Mock' and 'Mock'` | same #2763 normalisation | Give `response.usage` real ints, and assert the canonical output |

Each test now names the contract it enforces rather than having its number quietly loosened.

## A harness that was testing the wrong branch

Worth reading even if you skim the rest.

`test_gemini_adapter.py` stubs the SDK like this:

```python
mock_genai_pkg = MagicMock()
sys.modules["google.generativeai"] = mock_genai_pkg
```

A bare `MagicMock` auto-creates **any** attribute asked of it — including `Client`. So the adapter's probe:

```python
try:
    from google.generativeai import Client as _GenaiClient
    HAS_GEMINI_CLIENT = True
except (ImportError, AttributeError):
    HAS_GEMINI_CLIENT = False
```

came out `True` under test. Nine tests therefore exercised the per-instance `Client` branch while asserting against `configure` / `GenerativeModel` mocks that were **never touched**:

```
E   AssertionError: Expected 'configure' to be called once. Called 0 times.
E   assert <MagicMock name='mock.Client().models.get().start_chat()...'> == 'Hello there'
```

The real `google-generativeai` package exposes no `Client`, so production takes the legacy path — the branch the tests meant to cover. The autouse fixture now pins `HAS_GEMINI_CLIENT=False` explicitly, and `test_modern_client_path_is_preferred_when_available` covers the other branch on purpose, asserting the per-instance path never touches the process-global `configure()` (the #2756 footgun).

## What is deliberately left red

Nine tests still fail, all because they need **canonical Tools source**. They are listed here rather than skipped:

```
test_bitnet_adapter.py                3   prompt size / UTF-8 validation before Popen
test_gemini_adapter.py                2   two-argument _build_chat_session
test_{ollama,openai,anthropic}_adapter 3  empty current-message guard
test_ollama_adapter.py                1   typed httpx.ConnectError classification
```

All four regressions are fixed in **Tools#4574** and will pass here on the `vendor/ud-tools` bump — no source change needed in this repo.

## What changed and why

The first revision of this PR also edited five adapters under `src/shared/python/ai/adapters/`. CI was right to reject it:

```
FAILED tests/unit/repo_hygiene/test_tools_child_copy_contract.py::test_current_branch_does_not_edit_tools_child_copies
AssertionError: This branch directly modifies Tools-owned child copies in UpstreamDrift.
  Make each change in D-sorganization/Tools, merge or pin that canonical revision,
  and update only vendor/ud-tools here.
assert not ['ai/adapters/anthropic_adapter.py', 'ai/adapters/bitnet_adapter.py',
            'ai/adapters/gemini_adapter.py', 'ai/adapters/ollama_adapter.py',
            'ai/adapters/openai_adapter.py']
```

I had diagnosed those four regressions correctly and fixed them in the wrong repository. The guard exists precisely so that a downstream fix does not fork the canonical source — which, given that all four regressions were *caused* by a bulk consolidation, is the failure mode it is defending against. So the implementation work moved to Tools#4574, with 9 regression tests added there (Tools had none for any of the four, which is why they could be dropped without a red build upstream).

The `# noqa: BLE001` annotations from the previous revision are gone with the source changes; they are unnecessary now that this PR touches no adapter source.

## Verification

```
tests/unit/shared_python/ai/adapters/ + tests/ai/test_ai_adapters.py
  62 passed, 9 failed  (all 9 pending Tools#4574)

ruff check / ruff format --check   clean
suite-marker ratchet               passes (the one net-new test is marked)
```


dieterolson:
Independently verified the two load-bearing history claims. **Both hold exactly**, and I think this PR is more important than its title suggests — two of these are live production bugs, not test hygiene.

```
# BitNet prompt guard
$ git log -S'_MAX_PROMPT_BYTES' -- src/shared/python/ai/adapters/bitnet_adapter.py
0575fb4b8  2026-08-01  chore: consolidate overlapping UpstreamDrift PR backlog (#8322)   ← removed
0cddecfc6  2026-05-23  fix(ai): guard bitnet prompt size and encoding (#5997)            ← added
$ git show origin/main:...bitnet_adapter.py | grep -c _MAX_PROMPT_BYTES
0

# empty-current-message guard
$ git log -S'if current_message.strip():' -- src/shared/python/ai/adapters/ollama_adapter.py
3e09be404  2026-05-20  feat(sidekick): overhaul assistant settings, credentials, MCP …   ← removed
b5f391db7  2026-05-20  fix(sidekick): resolve ollama connection issues … (#5773)         ← added
$ git show origin/main:...ollama_adapter.py | grep -c 'current_message.strip()'
0
```

So a **chore/consolidation** PR deleted a security guard, and a **feature** PR deleted a correctness guard the same day it landed. Neither is in `main` today.

**The empty-message guard is a user-facing bug that has been live for three months.** Since 2026-05-20 every provider has appended an empty trailing user turn, and per your reading Gemini has additionally replayed the user's own message as history *while asking the model to answer a blank one*. That is degraded chat output in production, not a red test. It deserves to be called out in the PR title or at least the first paragraph — a reviewer skimming "repair the drifted tests" will not realise they are approving a production fix.

The **`MagicMock` auto-attribute finding is the best catch in the PR.** A bare `MagicMock` synthesising `Client` flipped `HAS_GEMINI_CLIENT` to `True` under test, so nine tests exercised a branch production never takes while asserting against mocks nothing touched. That failure mode is invisible in review and would have survived any amount of "just update the assertions" triage. Pinning it in an autouse fixture and covering the other branch deliberately is the right fix.

Also good: triaging all 19 individually instead of in bulk, landing on "both sides drifted, roughly evenly"; verifying the 2 residual failures are pre-existing by stashing and re-running; and confirming `ruff` finding-count is unchanged rather than just "clean".

### The systemic point

Your closing note — *"the four restored guards all trace to the same two merges … worth a look at what else those two dropped"* — is the most important sentence here, and it generalises beyond these two. From this sweep:

| Merge | What it silently removed |
| --- | --- |
| `3e09be404` | empty-message guard (4 adapters), Ollama typed error ladder |
| #8322 | BitNet prompt size/UTF-8 validation; `src/shared/python/tests` from `testpaths` (orphaning 8 files); 9 junk artifacts committed to repo root |
| #8746 | 1,627 DRY baseline fingerprints (wholesale `--write-baseline` inside a TypeScript memoisation PR) |

Three merges, each deleting protective code or protective *configuration*, none flagged at review — while the gates that should have caught them were skipping, false-green, or annotated "documented red". I'm filing that as its own issue rather than letting it live in a PR comment.

One blocker: `mergeable_state` is `dirty` — main has moved to `f2185fb7`, so this needs a conflict resolution before it can merge.

---

dieterolson:
Rebased onto `main` @ `7a24d4fdf`. Re-verified after the rebase: **9 failed, 62 passed** across `tests/unit/shared_python/ai/adapters/` and `tests/ai/test_ai_adapters.py`, unchanged from before — the ten test-side failures stay fixed, and the nine implementation-side ones are still the ones waiting on [Tools#4574](https://github.com/D-sorganization/Tools/pull/4574).

Worth restating what the split means for review, because the numbers alone look like a partial fix:

- The ten fixed here were **tests that were wrong**. Their assertions contradicted deliberate, issue-referenced implementation decisions (#5635, #2763, #3179), or — in the Gemini case — asserted against a mock branch the harness had silently switched away from. Nothing in `src/` needs to change for them.
- The nine still failing are **tests that were right**. Each describes behaviour that existed and was deleted by a bulk merge. Fixing them requires restoring the canonical implementation, which lives in Tools.

That is why this PR does not simply go green: making it green from inside UpstreamDrift would mean either editing Tools-owned source (which `test_tools_child_copy_contract.py` rejects, correctly) or weakening nine tests that are accurately describing a regression. Neither is acceptable, so the nine stay red and visible until the vendor bump.

Sequencing, for whoever lands these:

1. Tools#4574 merges.
2. `vendor/ud-tools` bumps here — no source change in this repo.
3. All nine flip green with no further edit to this branch.

This PR is independently mergeable before step 1; it strictly reduces the failure count and touches nothing the vendor bump will touch.

---

PR #8777, opened 2026-08-19
fix(deployment): restore False-returning hardware connect stubs (#8771)

dieterolson:
Part of #8771. Fixes `optional-stack-check (3.11)`, currently the only failing check on #8769 and #8774.

## The failure

The lane is otherwise green — 227 passed in biomechanics, 369 in robotics — and fails on exactly three tests:

```
FAILED test_spacemouse_input_without_hardware_backend    - NotImplementedError: Hardware driver not implemented (#8058).
FAILED test_vr_controller_input_without_hardware_backend - NotImplementedError: Hardware driver not implemented (#8058).
FAILED test_haptic_device_input_without_hardware_backend - NotImplementedError: Hardware driver not implemented (#8058).
```

Each does `assert not dev.connect()`. The three stubs in `src/deployment/teleoperation/devices.py` raise instead of returning falsy.

## Which side drifted

The implementation, and it is not a close call.

**#8322 introduced the raise, kept the tests, and deleted the reasoning along with the code.** Here is what it removed:

```diff
-        NOT IMPLEMENTED. There is no hardware driver behind this class, so the
-        honest answer is always "not connected" (#7360). The previous body
-        probed for ("pyspacemouse", "hid") and then returned ``False`` on
-        both arms of the branch, which made the stub look conditionally
-        functional -- installing the backend changed nothing. Tracked by #8058.
-
-        Returns:
-            Always ``False``.
+        NOT IMPLEMENTED. There is no hardware driver behind this class (#8058).
+
+        Raises:
+            NotImplementedError: Always raised since hardware backend is missing.
-        self._is_connected = False
-        return False
+        raise NotImplementedError("Hardware driver not implemented (#8058).")
```

Three times, once per device. So `False` was not an accident anyone forgot to update — it was a decision with an issue behind it (#7360), specifically about not making a stub *look* conditionally functional. It was replaced by a raise with no argument offered anywhere in the diff or the commit message.

**Returning `False` is also what the rest of the subsystem does.** Three independent confirmations, all currently passing:

| Evidence | Behaviour |
| --- | --- |
| `BaseInputDevice.connect` — the direct parent of all three | `self._is_connected = False; return False` |
| `test_controller_stubs.py::TestHardwareConnectStubs` (ROS2, UDP) | `assert result is False, "Unimplemented ROS2 connect must return False"` |
| `test_devices.py::test_base_input_device` | `assert not dev.connect()` on the base class |

So the three overrides were inconsistent with their own parent class, with the sibling protocol stubs, and with their tests simultaneously. The only thing they were consistent with was #8322.

## Why raising is the wrong contract here regardless of history

The whole file is built around graceful degradation. `_assert_disconnected_device_contract` — the shared helper all three tests call — expects a device to be constructible, probe-able, and *then* to refuse state queries with a typed `StateError`:

```python
with pytest.raises(StateError, match=f"{type(dev).__name__} is not connected"):
    dev.get_pose()
```

That contract still works and is still asserted. Raising from `connect()` breaks the probe step that precedes it, so a caller written the obvious way —

```python
if not device.connect():
    fall_back_to_keyboard()
```

— gets an exception instead of the fallback. `KeyboardMouseInput.connect()`, the one device that *is* implemented, returns `True`, which is exactly the branch that becomes unreachable.

The "#8058 is not implemented" signal is not lost by this change: it stays in the docstring, and `_require_connected` still raises `StateError` on every method that would need real hardware.

## The fix

Restores the pre-#8322 bodies verbatim, plus a note on *why* `False` rather than a raise, so the next merge has to argue with it rather than silently reverse it again:

```
Returning ``False`` rather than raising is the deliberate contract: it
matches ``BaseInputDevice.connect``, the ROS2/UDP controller stubs, and
the graceful-degradation path every caller and test relies on. #8322
replaced it with ``raise NotImplementedError`` and kept the tests, which
has redded ``optional-stack-check`` ever since (#8771).
```

**No test was changed, skipped, or deleted.**

## Verification

All three `optional-stack-check` unit targets, run locally with the lane's marker expression:

```
tests/unit/deployment      33 passed          (was 3 failed / 30 passed)
tests/unit/robotics       369 passed
tests/unit/biomechanics   227 passed, 1 failed
```

The one biomechanics failure is `test_plot_grf_and_com_3d`:

```
ImportError: Cannot load backend 'QtAgg' which requires the 'qt' interactive
framework, as 'headless' is currently running
```

Reproduced identically on pristine `main` with this branch stashed, and the lane runs its pytest under `xvfb-run` where it passes (the job log shows biomechanics green). Environmental to my sandbox, not a regression, not touched here.

`ruff check` and `ruff format --check` clean on the changed file.

## Note

This is the third place in the #8771 sweep where #8322 deleted hardening and kept its tests — the others being `BitnetAdapter`'s prompt validation and Ollama's typed error ladder, both in #8775. That merge is worth a systematic audit; there are likely more.

---


dieterolson:
CI on the first commit found the other half of this, and it is worth reading before reviewing — it makes the case considerably stronger than the PR description originally claimed.

## `unit-test-gate` failed with the mirror image of the bug

```
FAILED tests/deployment/test_teleoperation.py::test_spacemouse_input                    - DID NOT RAISE NotImplementedError
FAILED tests/deployment/test_teleoperation.py::test_vr_controller_input                 - DID NOT RAISE NotImplementedError
FAILED tests/deployment/test_teleoperation.py::test_haptic_device_input                 - DID NOT RAISE NotImplementedError
FAILED tests/deployment/wave5_deployment/test_teleoperation.py::TestInputDevices::test_spacemouse_reports_no_hardware_and_refuses_state_queries
FAILED tests/deployment/wave5_deployment/test_teleoperation.py::TestInputDevices::test_vr_controller_reports_no_hardware_and_refuses_state_queries
FAILED tests/deployment/wave5_deployment/test_teleoperation.py::TestInputDevices::test_haptic_reports_no_hardware_and_refuses_state_queries
6 failed, 11343 passed, 604 skipped
```

**Three test files cover these three stubs, and `main` currently asserts both contracts at once.** #8322 rewrote two of them to match its new raise, and missed the third:

| File | #8322 | contract on `main` today |
| --- | --- | --- |
| `tests/deployment/test_teleoperation.py` | rewritten (3 assertions) | raises |
| `tests/deployment/wave5_deployment/test_teleoperation.py` | rewritten (3 assertions) | raises |
| `tests/unit/deployment/test_devices.py` | **missed** (3 assertions) | returns `False` |

That single omission is the whole story of this failure. `unit-test-gate` runs `tests/deployment/**` and is satisfied by the raise; `optional-stack-check` runs `tests/unit/deployment` and is satisfied by `False`. Two lanes have been enforcing opposite halves of a contract that was cut in half on 2026-08-01, and only one of them was red, so it read as an isolated three-test problem rather than a subsystem-wide inconsistency.

## The pre-#8322 state was coherent

Checked against `0575fb4b8^`. All three files asserted `False`:

```
tests/deployment/test_teleoperation.py:49            assert not mouse.connect()
tests/deployment/wave5_deployment/...py:222          assert d.connect() is False
tests/unit/deployment/test_devices.py:49             assert not dev.connect()
```

So this is not "two tests say one thing, one says another, pick a winner". There was one contract, documented under #7360, asserted identically in three places and matching `BaseInputDevice.connect` and the ROS2/UDP stubs. #8322 reversed it and updated two thirds of its own evidence.

## What changed in the second commit

`ae590392c` restores those six assertions to their pre-#8322 text. Both files are byte-identical to `0575fb4b8^` for these blocks — `tests/deployment/test_teleoperation.py` is identical outright, and the wave5 file differs only by a `pytestmark = pytest.mark.unit` line added later by unrelated work.

I want to be explicit that I am changing tests here, because the brief for #8771 says not to. The distinction I am drawing: these six assertions are not independent evidence of a contract, they are part of the same unreviewed change that broke it, and reverting the implementation without reverting them would leave the split in place merely facing the other way. Nothing is skipped, deleted, or `xfail`ed, and nothing is weakened — `assert d.connect() is False` is a *stricter* assertion than `pytest.raises`, because it pins a specific return value rather than any raise.

If a reviewer prefers the raise contract, the fix is the mirror image and equally mechanical: rewrite the three assertions in `tests/unit/deployment/test_devices.py`, and separately reconcile `BaseInputDevice.connect` and the ROS2/UDP stubs, which return `False` and have their own passing tests. What is not tenable is leaving both contracts asserted, which is the state `main` is in now.

## Verification

```
tests/unit/deployment + tests/deployment    288 passed
ruff check / ruff format --check            clean
```

First commit's CI, for the record: `code-quality`, `security-scans`, `repo-structure-gates`, `tests`, `rust-wheel-parity` all **success**; only `unit-test-gate` failed, on exactly these six.

---

dieterolson:
Two more CI results, one of which is a finding about the runner fleet rather than about this PR.

## `optional-stack-check (3.11)` failed without running a test — self-hosted runner defect

This is the gate the PR exists to fix, so it matters that its failure here is **not** the tests. The job died in setup after 33 seconds:

```
Runner name: 'd-sorg-local-ControlTower-SSD-4'
...
Version 3.11 was not found in the local cache
Version 3.11 is available for downloading
Execute installation script
Check if Python hostedtoolcache folder exist...
Create Python 3.11.16 folder
##[error]mkdir: cannot create directory
  '/home/dieterolson/actions-runners/runner-4/_work/_tool/Python/3.11.16': Permission denied
##[error]The process '/usr/bin/bash' failed with exit code 1
```

`actions/setup-python` cannot write to `_work/_tool` on that runner, so no Python was installed and no test ran. Nothing downstream of that step executed — the artifact it uploaded contains only the empty skip-report files.

**This is an environment problem on `d-sorg-local-ControlTower-SSD-4`, not a code one**, and it will fail identically for any PR whose `optional-stack-check` job lands on runner-4. Worth fixing at the fleet level (`chown`/`chmod` the tool cache, or pre-seed 3.11.16 in `_work/_tool`), because until then this required check is a coin flip on runner assignment. Re-running the job should pick a different runner.

For the record, the three tests this PR targets, run locally with the lane's own marker expression:

```
tests/unit/deployment      33 passed     (was 3 failed / 30 passed)
tests/unit/robotics       369 passed
tests/unit/biomechanics   227 passed, 1 failed  (matplotlib QtAgg; needs the lane's xvfb-run)
```

## `guard` (Stub Introduction Guard) — my own fault, fixed in `a84115073`

Legitimate catch. The guard scans added lines for `raise NotImplementedError` and requires an issue reference **on the same line**. My docstring narrates what #8322 did:

```
the graceful-degradation path every caller and test relies on. #8322
replaced it with ``raise NotImplementedError`` and kept the tests, which
has redded ``optional-stack-check`` ever since (#8771).
```

`(#8771)` wrapped onto the following line, so the guard saw an untracked stub introduction. Reproduced the scan locally against the PR diff — three hits, one per device — and rewrapped so the reference sits on the line that mentions the pattern. Same text, same meaning, zero violations now:

```
$ python3 <stub-guard scan> pr.diff
violations: 0
```

A slightly amusing outcome for a PR whose whole subject is a stub that should not raise.

## Current state

```
tests/unit/deployment + tests/deployment    288 passed
ruff check / ruff format --check            clean
stub-introduction guard scan                0 violations
```

`unit-test-gate` was still in progress at the time of writing; the six failures it reported on the first commit are addressed by `ae590392c`.

---

dieterolson:
Independent confirmation that this fix is complete and correct, from a run on **another PR's branch** — plus a fleet problem that needs an admin.

## The lane's only failures are the three tests this PR fixes

`optional-stack-check (3.11)` ran to completion on #8774's branch (which does **not** contain this change) on runner `d-sorg-local-Oglaptop-7`:

```
## Optional-stack unit target: tests/unit/biomechanics
================== 210 passed, 1 skipped, 1 warning in 3.47s ===================
## Optional-stack unit target: tests/unit/deployment
=================== 3 failed, 30 passed, 1 warning in 2.23s ====================
##[error]Unit tests failed in the optional-stack lane with pytest exit code 1.
```

Plus `659 passed, 1 xfailed` and `7 passed` in the earlier API and Pinocchio targets. So the entire lane is green except `tests/unit/deployment`, and within that, except the three `connect()` tests. `30 passed` there becomes `33 passed` with this branch — which is exactly the local result:

```
tests/unit/deployment      33 passed
```

Nothing else in the lane is affected by the change, which is what you'd hope from a three-method revert.

## The remaining red on this PR is a broken tool cache on part of the runner fleet

`optional-stack-check` has now failed twice on this PR without executing a single test, on two different self-hosted runners:

| run | runner | failed step | outcome |
| --- | --- | --- | --- |
| `32207742515` | `d-sorg-local-ControlTower-SSD-4` | `actions/setup-python` | `mkdir: Permission denied` |
| `32208288970` | `d-sorg-local-ControlTower-SSD-6` | `actions/setup-python` | same |
| `32207913135` (#8774) | `d-sorg-local-Oglaptop-7` | — | ran fine, 876 tests |

```
Version 3.11 was not found in the local cache
Version 3.11 is available for downloading
Execute installation script
Check if Python hostedtoolcache folder exist...
Create Python 3.11.16 folder
##[error]mkdir: cannot create directory
  '/home/dieterolson/actions-runners/runner-N/_work/_tool/Python/3.11.16': Permission denied
```

The mechanism looks like this: a **new patch release, 3.11.16**, is now what `setup-python` resolves `python-version: "3.11"` to. The `ControlTower-SSD-*` runners have an older 3.11.x in their tool cache but no write access to `_work/_tool`, so the action cannot install the new patch and dies. `Oglaptop-7` either has 3.11.16 already or a writable cache.

**This is not specific to this PR** — any job on those runners that uses `setup-python` with an uncached version will fail the same way, and `optional-stack-check` is a required check. Until it is fixed, whether a PR passes that gate is decided by runner assignment. Two candidate remedies for whoever owns the fleet:

- `chown -R` the runner user over `_work/_tool` on the `ControlTower-SSD-*` runners, or
- pre-seed `3.11.16` into their tool cache, or
- pin `python-version: "3.11.x"` in the workflow to an already-cached patch, which is a workaround rather than a fix and will recur.

Re-running the job is the short-term unblock; it will eventually land on a healthy runner.

Happy to file this separately if it should not live on this PR — it is unrelated to the change, it just happens to be what is standing between this PR and a green check.

---

PR #8780, opened 2026-08-19
Bind the Dispersion and Trends tabs to corpus data

dieterolson:
## Summary

Two tabs of the Launch Monitor Analytics workbench have been inert against corpus data since the corpus was wired up in #8757/#8769 — **Dispersion** had no lateral coordinate to bind to, and **Trends** had no time column. The data authority has now extracted both (its issues #18 and #19); this is the consumer half.

- `lateral_carry_yd` → **`lateral_carry`** (yards → metres) and `flight_time_s` → **`flight_time`** join `CORPUS_COLUMN_MAP`.
- `captured_at` is carried through as an optional identity column — it is exactly what `analyze_trend` binds to.
- Column selection is filtered against the dataset's actual schema, so a corpus pinned before these columns existed still loads rather than failing the whole read on a missing name.

## Verified by actually running the tabs

Driven headless against an authorized checkout, not inferred from the data being present:

```
DISPERSION: Dispersion complete for 1 group(s).     20,099 shots, 95% ellipse
TRENDS:     Trend slope=-0.01793/day; 3 candidate change point(s).   8,488 shots
```

The dispersion ellipse renders with `center_forward` 136.8, `center_lateral` 0.56, `ellipse_major` 308.7 — a centred, plausibly-shaped distribution rather than a degenerate one.

Corpus-wide the new columns arrive as: `lateral_carry` 20,099 rows (median |10.0| m), `flight_time` 223,956 rows (max 9.67 s), `captured_at` 8,488 rows across 13 sources.

## One bug worth naming

`load_private_corpus` already used the local name `available` for the **set of source ids** inside the `sources is not None` branch. My first pass introduced a second `available` for column names, which the sources branch then clobbered — so the column filter matched nothing and returned an empty frame. Caught by the existing `test_source_and_metric_selection`; the column set is now `available_columns`. Worth flagging because the two sets are both plausibly called "available" and the collision is silent.

## Test plan

- [x] `pytest tests/unit/launch_monitor/` — all pass (2 new)
- [x] New tests cover canonical conversion of all three columns **and** that a corpus predating them still loads
- [x] `ruff check` / `ruff format --check` clean on changed files
- [x] Live workbench run producing the Dispersion ellipse and Trends slope quoted above
- [x] SPEC.md updated (1.0.544) per the spec-check policy

Depends on data-authority PR #20 for the columns to be present; degrades cleanly to today's behaviour until the lock file is bumped.

---

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -485,7 +485,7 @@ inventory and reopen adjudication until every new candidate is reviewed.
 | **License**             | MIT                                                |
 | **Current Version**     | 2.1.1                                              |
 
-| **Spec Version** | 1.0.543 |
+| **Spec Version** | 1.0.544 |
 | **Last Spec Update** | 2026-08-16 |
 
 ## 2. Purpose & Mission
@@ -871,6 +871,14 @@ study,vandv,provenance,units}`, with subpackages re-exported by name and a
   tabs remain inert for corpus data because the corpus carries no capture
   timestamp or lateral carry; both gaps are tracked as data-authority issues
   #18 and #19 and are recoverable from retained native fields.
+- **2026-08-19** - Bound the Launch Monitor Analytics Dispersion and Trends
+  tabs to corpus data. The data authority now extracts lateral carry, flight
+  time, and a capture date, and `launch_monitor.corpus` maps them onto the
+  canonical `lateral_carry` (m), `flight_time` (s), and `captured_at` fields.
+  Both tabs were previously inert against the corpus for want of a lateral
+  coordinate and a time column; they now run over 20,099 and 8,488 shots
+  respectively. Column selection is filtered against the dataset schema, so a
+  corpus pinned before those columns exist still loads.
 - **2026-08-18** - Connected Launch Monitor Analytics to the private shot
   corpus. `launch_monitor.corpus.load_private_corpus()` reads the data
   authority's source-partitioned Parquet corpus (261,666 shots across 27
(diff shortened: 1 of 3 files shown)


dieterolson:
Status on the two red checks.

**`repo-structure-gates` — mine, fixed in `be553add9`.** Legitimate failure caused by this diff:

```
src/shared/python/launch_monitor/corpus.py:70:
  load_private_corpus exceeds function-lines budget (109 > 100)
```

Adding the schema-tolerant column selection pushed it over. Split into four named helpers — `_selected_column_map`, `_source_filter`, `_canonicalize_metrics`, `_apply_identity` — rather than filing a budget exception, since the function was doing four separable things. `check_architecture_budget.py` now passes locally, `ruff check`/`format --check` clean, and the launch-monitor unit, UI, and API suites all still pass.

The split also removes a latent hazard: the source-partition scan used a local named `available`, one character from the `available_columns` set introduced in this PR. Moving it into `_source_filter` makes the collision impossible.

**`optional-stack-check (3.11)` — pre-existing on `main`, not from this diff.** Three tests in `tests/unit/deployment/test_devices.py`:

```
test_spacemouse_input_without_hardware_backend
test_vr_controller_input_without_hardware_backend
test_haptic_device_input_without_hardware_backend
```

All fail the same way — `connect()` raises `NotImplementedError` where the test asserts it returns falsy. Confirmed to fail identically on `origin/main` with this branch stashed, and it is teleoperation device drivers, untouched by this PR. Tracked under #8771; will re-run once the base branch recovers.

---

PR #8781, opened 2026-08-19
fix(docker): make the modular builder slice self-sufficient (#8771)

dieterolson:
Part of #8771. Fixes all five `profile-size-matrix` checks — `slim`, `standard`, `biomech`, `research`, `full` — which are red on `main` and were not in the issue's original inventory.

**Reshaped** after CI: the first revision fixed the right bug in the wrong file. See "What changed and why" at the bottom.

## The failure

All five die at the same Docker build step, before any image is produced:

```
#18 [builder 12/22] RUN set -eux; ... install_features.py --profile "slim" --dry-run
#18 0.075 [install_features] resolved features: api, pendulum, mujoco
#18 0.075   File "/workspace/scripts/docker/install_features.py", line 166, in _feature_install_argv
#18 0.075     from src.shared.python.feature_registry.features import get_feature
#18 0.075   File "/workspace/src/shared/python/__init__.py", line 32, in <module>
#18 0.075     from . import ai as ai
#18 0.075 ImportError: cannot import name 'ai' from partially initialized module 'src.shared.python'
ERROR: failed to build: failed to solve
```

## Root cause

`Dockerfile.modular`'s builder stage copies a **deliberate minimal slice**:

```dockerfile
COPY src/__init__.py                        ./src/__init__.py
COPY src/shared/__init__.py                 ./src/shared/__init__.py
COPY src/shared/python/__init__.py          ./src/shared/python/__init__.py
COPY src/shared/python/engine_core/         ./src/shared/python/engine_core/
COPY src/shared/python/feature_registry/    ./src/shared/python/feature_registry/
RUN ... install_features.py --profile "$PROFILE" --dry-run
```

with the reason a few lines below: *"The full source tree is COPYd in the runtime stage so layer caches"* are not invalidated by every source change. Resolving which pip packages a profile needs must not require the whole repo.

`install_features.py` reached the registry through a **dotted import**, and importing a submodule executes every parent package's `__init__` first. So the slice was only ever self-sufficient by luck — it depended on `src/shared/python/__init__.py` importing nothing outside the slice. When that `__init__` grew an eager `from . import ai`, every modular image build broke. (The traceback's *"most likely due to a circular import"* is CPython guessing; the module is simply absent.)

## The fix

Load `features.py` from its path, with no parent package:

```python
spec = importlib.util.spec_from_file_location("_install_features_registry", features_path)
module = importlib.util.module_from_spec(spec)
sys.modules[spec.name] = module      # see below
spec.loader.exec_module(module)
```

This is what the comment above that code **always claimed** to do:

> To avoid running the registry's import-time DbC validator … we read the features module **in isolation** by adding the repo to sys.path.

Adding the repo to `sys.path` never achieved isolation — the dotted import still ran all three `__init__`s. Now it genuinely does, which makes the builder slice self-sufficient *regardless of what any package `__init__` imports in future*. `features.py` depends only on `dataclasses` and `typing`.

**One subtlety worth flagging for review:** the module must be registered in `sys.modules` *before* `exec_module`, because `@dataclass` resolves `cls.__module__` through `sys.modules` while processing the class body. Without it you get:

```
File "/usr/lib/python3.11/dataclasses.py", line 712, in _is_type
  ns = sys.modules.get(cls.__module__).__dict__
AttributeError: 'NoneType' object has no attribute '__dict__'
```

I hit exactly that on the first attempt. It is cleaned up on failure so a failed load leaves no half-initialised entry behind.

## Verification

Reconstructed the builder slice — exactly the paths the Dockerfile copies, nothing else — with `src/shared/python/__init__.py` left **exactly as it is on `main`**, eager import and all:

| | result |
| --- | --- |
| `origin/main` | `ImportError: cannot import name 'ai'`, exit 1 |
| this branch | all five profiles exit 0 |

```
slim     -> exit=0  (api, pendulum, mujoco)
standard -> exit=0  (… + pinocchio)
biomech  -> exit=0  (… + opensim, myosuite)
research -> exit=0  (… + drake, pose-mediapipe)
full     -> exit=0  (… + chrono)
```

Full-tree behaviour unchanged — `--profile slim/full` and `--features mujoco` resolve identically, and an unknown profile still fails loudly with the known list:

```
$ install_features.py --profile nonexistent --dry-run
unknown profile 'nonexistent'; known: biomech, full, gpu-training, research, slim, standard
```

```
tests/unit/repo_hygiene/test_tools_child_copy_contract.py   9 passed, 3 skipped
ruff check / ruff format --check                            clean
```

## What changed and why

The first revision restored the PEP 562 lazy `__getattr__` in `src/shared/python/__init__.py`. That was the right diagnosis and the wrong file — `unit-test-gate` caught it:

```
FAILED tests/unit/repo_hygiene/test_tools_child_copy_contract.py::test_current_branch_does_not_edit_tools_child_copies
  This branch directly modifies Tools-owned child copies in UpstreamDrift.
  Make each change in D-sorganization/Tools, merge or pin that canonical revision,
  and update only vendor/ud-tools here.
```

**And a vendor bump would not have fixed it either.** The builder stage copies `src/shared/python/__init__.py` from *this repo's own tree*, not from `vendor/ud-tools/` — so pinning a fixed Tools revision leaves the file the Dockerfile actually reads unchanged. I checked the `COPY` list specifically for this.

That leaves the fix having to live in UpstreamDrift-owned code, and `scripts/docker/` is exactly that. It is also the better fix: it removes the coupling permanently instead of restoring a property that has to be maintained by convention. The eager import in Tools remains a latent bug worth fixing upstream — nothing else currently depends on that `__init__` being lazy, but it is a trap — and it is no longer what breaks these builds.

Nothing currently *tests* that the builder slice is self-sufficient, which is why this broke silently for two and a half weeks. A cheap guard would be a unit test that reconstructs the slice from the `COPY` lines and runs the validator against it; I have not added one here, since it wants a home and a convention decision rather than being tacked onto a fix.

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -485,7 +485,7 @@ inventory and reopen adjudication until every new candidate is reviewed.
 | **License**             | MIT                                                |
 | **Current Version**     | 2.1.1                                              |
 
-| **Spec Version** | 1.0.548 |
+| **Spec Version** | 1.0.549 |
 | **Last Spec Update** | 2026-08-19 |
 
 ## 2. Purpose & Mission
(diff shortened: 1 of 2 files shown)


dieterolson:
## Correction: this did not fix all five `profile-size-matrix` checks, and I should not have said it did

The PR body claims "Fixes all five `profile-size-matrix` checks". **That is wrong**, and the error is mine — I verified one of two blockers and generalised.

### What the change did do

It works, and it is necessary. Comparing builder progress on `slim`:

```
before:  #18 [builder 12/22]  ImportError: cannot import name 'ai'    <- died here
after:   #18 [builder 12/22]  ok
         #29 [builder 22/22]  RuntimeError ...                        <- dies here
```

Step 12 is the `--dry-run` profile validation. It passes now, and the build gets ten steps further.

### The second blocker, which was masked behind the first

Step 22 is the **real** install — `install_features.py --profile slim` without `--dry-run` — which runs `pip install .`, which runs `build_hooks.initialize`:

```
#29 7.168   File "/workspace/build_hooks.py", line 361, in initialize
#29 7.168   File "/workspace/build_hooks.py", line 294, in _register_tools_packages
#29 7.168     raise RuntimeError(message)
#29 7.168 RuntimeError: Pinned Tools package roots are required for release/CI builds;
#29 7.168   missing: shared, sidekick, chat, python/src/utils, contracts.py
ERROR: failed to build: failed to solve
```

`_register_tools_packages` raises when the canonical Tools roots are absent and `version != "editable" or _env_flag("CI")`. In the builder stage `CI` is set and `vendor/ud-tools` is **never copied into that stage at all** — `Dockerfile.modular` copies it in neither the builder nor the runtime stage, and `docker-size-gates.yml`'s `actions/checkout` does not request submodules. So the raise is unconditional there.

This has been failing the whole time; it was simply unreachable behind the `ImportError`.

### Why my verification missed it

I reconstructed the builder *slice* — the five paths the `COPY` lines list — and ran `install_features.py --dry-run` against it. That exercises **step 12 only**. Step 22 is a real `pip install .` and only surfaces in an actual `docker build`, which I did not run. The reconstruction was a good way to isolate the first bug and a bad way to claim the checks were fixed, and I should have said "fixes the `ImportError` that blocks the builder at step 12" rather than naming the checks.

### What remains

Two candidate fixes, neither of which I have implemented or tested:

- **(a) Copy `vendor/ud-tools/` into the builder stage** before the install step, and have `docker-size-gates.yml` check out submodules. Keeps the fail-loud guard the Dockerfile comment at line 90 explicitly wants ("...which is the correct fail-loud behavior"). Costs build context size and cache locality.
- **(b) Make `_register_tools_packages` degrade rather than raise** when the pinned roots are out of scope for the stage. Smaller diff, but it weakens a guard that exists to stop release artefacts shipping without vendored Tools — I would not do this without an owner's sign-off, and it is the kind of change #8771 exists to discourage.

(a) looks right to me, but it is a real design decision about what the builder stage is allowed to depend on, and it deserves its own PR rather than being appended here.

### Also seen, separately

The post-merge run on `main` @ `eb7ce2da7` failed *before* reaching either blocker:

```
##[error]ERROR: docker endpoint for "default" not found
```

on `d-sorg-local-Desktop-7` — a Docker/buildx setup failure, third distinct cause, same self-hosted fleet as the `setup-python` tool-cache failures. So `profile-size-matrix` currently has one real remaining code blocker plus at least one environmental one.

Leaving the record straight rather than letting a merged PR claim a green it did not deliver.

---

Issue #8733, opened 2026-08-17
F1 MPM: remainder of the tier (ball body, whole-shot march, plastic-limit verification, F2 calibration)


dieterolson:
## Context

PR for the F1 solver core landed the 2-D plane-strain MPM solver at
`src/bunkershot3d/solvers/mpm/` with its code verification (ADR-0033).
That PR was deliberately scoped to **solver core plus verification**.
This issue records what was left out, so the gaps are tracked rather
than discovered.

None of the items below are defects in what shipped. They are the
remainder of the tier.

## 1. The ball as a plane-strain body

ADR-0033 decides that "the ball becomes a body inside F1, as a rigid
circular section in the plane-strain plane, coupled to the sand by the
same traction integration used for the sole", and states this is a
modelling change decided in the ADR rather than left for #8712 to
discover.

`RigidSection` already carries everything needed (convex polygon, rigid
velocity field, swept collision, exact momentum ledger), so a circular
section is a polygonal approximation plus a second body in the step
loop. Two things must travel with it:

- in plane strain the ball is an **infinite cylinder, not a sphere**, so
  flux onto it is per unit width and its geometry is wrong in the third
  dimension;
- the below-equator / face-side split #8712 asks for is in-plane and can
  be addressed **qualitatively**; any heel-toe or lateral distribution
  over the ball is already refused by
  `RefusedQuantity.OUT_OF_PLANE`.

Ball launch stays on F0's #8657 momentum-transfer path;
`RefusedQuantity.BALL_LAUNCH` already raises for it.

## 2. Multi-body contact in one step

`_advance` takes a single optional `RigidSection`. Adding the ball means
a sequence of bodies with one momentum ledger each. The contact
projection is order-dependent when two bodies overlap the same node, so
the ordering has to be chosen and stated rather than left to whichever
loop order the implementation happens to use.

## 3. F1 time-marching over a whole shot

F1 currently answers one `IntrusionState` by building its own approach
history: the body is reversed along its velocity until clear of the bed,
then driven back to the queried pose at constant velocity. That is a
declared modelling assumption and it is what makes an F1 answer
comparable to F0's memoryless one.

A genuine F1 shot would instead march the head's real trajectory once,
in the shape of `solvers/shot.py::simulate_shot`, and read the wrench
history off a single continuous solve. That is strictly better and
strictly more expensive, and it is the right basis for the
peak-load-timing comparison ADR-0033 marks quotable-with-tier.

[...]

Issue #8712, opened 2026-08-16
B4: Ball-sand interaction — what actually reaches the ball, by region


dieterolson:
Resolve and visualise the sand that reaches the ball: velocity and mass flux by region of the ball's surface, over the contact window.

## This requires a modelling change, not just a view

At F0 the ball is a downstream result — `compute_bunker_launch` derives launch from solver impulse (#8657), and nothing strikes the ball. There is no "sand impacting the ball" to visualise, even in principle. Answering the question requires the ball to be a body inside the sand-solving tier chosen in B1, so that grains or continuum actually reach it.

That change is also the most likely route to fixing the backwards ball-speed artefact, where fluffy and plugged lies currently produce *higher* ball speed than firm sand — a transfer model with fixed efficiency cannot represent a lie-dependent splash, but a resolved one can.

## Scope

- Sand velocity and mass flux at the ball surface, by region, over time.
- Comparison of regions — below-equator versus face-side — since that distribution is what determines launch and spin.
- Explicit labelling of tier and validity; if the tier is an illustrative grain proxy, the ball-side numbers are illustrative too and must say so.

Blocked by B1. Related to #8657 and the ball-speed artefact issue.

Part of #8699.

Issue #8657, opened 2026-08-15
Derive bunker ball launch from the solver's delivered impulse, not a hard-coded 12 cm sweep length


dieterolson:
## Carry is the weakest number in the tool, and it drives the headline objective

`bunkershot3d.ball.splash` derives ball launch from a hard-coded box of displaced sand:

```python
# src/bunkershot3d/ball/splash.py:165-167
# Volume = sole_length * entry_depth * sweep_length (approximately)
sweep_length = 0.12  # ~12cm of sand contact (typical splash shot)
displaced_volume = sole_length_m * entry_depth_m * sweep_length
```

plus a flat transfer efficiency at line 190. Ball speed comes out **linear in entry depth**.

This is not fabricated — it is a reasonable first-cut estimate, and it was written before the
F0 solver existed. But it now has three problems, and they compound:

1. **It drives `playability_window_area`**, the primary scalar objective of the whole tool
   (#8614). Everything the workbench ranks rests on it.
2. **It duplicates, worse, a quantity the solver now computes.** `sweep_length = 0.12` is a
   guess at the sand contact length. The F0 solver returns contact duration and entry/exit
   speeds, and the metrics layer returns divot length, depth and mass directly. We are
   estimating with a magic number something we measure two modules away — a DRY violation
   where the duplicated copy is the less accurate one.
3. **It does not declare itself.** Every other borrowed or uncalibrated quantity in this
   package carries provenance: sand constants are `BORROWED_ANALOGUE`, solver coefficients
   report `measured_constants() == ()`, and out-of-envelope solver queries refuse outright.
   Carry is the one headline number that presents itself with the same confidence as a
   verified one. That is inconsistent with the package's own honesty rules (#7999, #8616).

## The fix

**Derive ball launch from the momentum the solver actually delivered to the sand**, not from
a fitted line through entry depth.

The solver already produces what is needed:

- `impulse_n_s` — the impulse delivered, which is the physically meaningful driver
- contact duration, entry and exit speed
- divot length, depth and **mass** from `metrics.divot` — i.e. the real displaced volume,
  rather than `sole_length * depth * 0.12`

So: replace the box estimate with the computed divot mass and the solver's impulse, keep the
transfer efficiency as an explicitly-named, explicitly-uncalibrated parameter, and delete
`sweep_length` entirely.

Then **give the result provenance and a validity verdict** in the same shape as the solver's,
so a carry figure can never be quoted as if it were measured. Per #8616, no published data
[...]

PR #9172, opened 2026-08-29
feat(bunkershot3d): F1 ball body, multi-body contact and whole-shot marching (#8733 sections 1-3)

dieterolson:
Implements **sections 1, 2 and 3** of #8733 (epic #8699): the F1 MPM tier's ball
body, multi-body contact, and whole-shot marching. Sections 4 (verification the
current suite does not reach) and 6 (calibration toward F2) are untouched, so the
issue stays open.

`verification.py` and `test_verification.py` are not touched.

## §1 — The ball as a plane-strain body (`solvers/mpm/ball.py`)

ADR-0033 had already decided this: the ball is a rigid circular section in the
plane-strain plane, coupled by the same traction integration as the sole.
`RigidSection` already carried the convex polygon, the rigid velocity field, the
swept collision and the exact momentum ledger, so a circle is a polygonal
approximation plus a second body.

The polygon is **equal-area**, not inscribed or circumscribed. Either of those
biases the displaced area systematically and the bias survives grid refinement,
so it would read as a quiet offset in the flux onto the ball rather than as
noise. Scaling the circumradius by `sqrt(2 pi / (n sin(2 pi / n)))` makes the
section area exactly `pi R^2` at every facet count, so the polygon crosses the
circle instead of sitting inside or outside it. `n_facets_for_cell_size` then
picks a facet count whose chord fits inside one cell — the only scale at which
the sand can see the flat spots.

The two facts ADR-0033 requires travel **in the API**, not in a comment:

- **Infinite cylinder, not a sphere.** `line_mass_kg_per_m` is named for the
  per-unit-width quantity it is, and `sphere_mass_kg` *raises* rather than
  returning `rho (4/3) pi R^3`, because a number that can be returned will be
  quoted. `PLANE_STRAIN_BALL_NOTE` is carried into `BallContactSplit.summary()`.
- **The below-equator / face-side split is in-plane and qualitative.**
  `BallContactSplit` reports fractions, flags itself `is_qualitative`, and its
  halves sum to the total exactly. `BallSection.heel_toe_split`,
  `BallSection.lateral_distribution` and `BallContactSplit.heel_toe_fraction` all
  raise `RefusedQuantity.OUT_OF_PLANE`.

`BallSection.launch_velocity_m_s` raises `RefusedQuantity.BALL_LAUNCH`. Ball
launch stays on F0's #8657 momentum-transfer path.

## §2 — Multi-body contact in one step (`solvers/mpm/contact.py`)

The grid projection writes a velocity-level constraint straight onto the node, so
when two bodies share a node the **last** projection wins and the earlier body's
non-penetration may be left violated. The order is therefore chosen and stated
rather than inherited from a loop:

> **Slowest first, fastest last, ties in the caller's order.**

Two reasons, in the order they matter:

1. **The fastest body is the one that can tunnel.** Everything in `body.py` about
   not letting sand through the club — the CFL bound on body travel, the
   swept-node test, the pushout backstop — scales with body speed. Applying the
   fastest body last is what makes its non-penetration exact, and it is the body
   for which "exact" is worth paying for.
2. **The order is a function of state, not of arguments.** Two callers who pass
   the same bodies in different sequences get the same answer.

Three tests carry the choice: that the order genuinely bites (fast-first and
fast-last differ), that the fastest body's constraint holds exactly, and that
what the choice *costs* is the slow body under-collecting — never sand passing
through the fast one.

**The momentum ledger is unaffected by the order.** Each body records
`m_i (v_i^after - v_i^before)` at its own stage, so the stages telescope. A
two-body march of a 600-particle bed with every wall `FREE` — gravity and the
bodies being the only momentum sources — closes
`delta p = sum_b J_b + M g dt n` to a residual of **1.95e-16 kg m/s** against a
2.6e-2 change (**4.3e-15 relative**), with the club delivering
`(2.631e-2, -3.249e-2)` and the ball `(-9.25e-7, -3.760e-3)` N s per metre of
width, and zero pushouts.

An iterated projection would remove the order dependence entirely. It is not done
here, and the reason is written down: the bodies overlap at a handful of nodes
for a handful of steps, and an unconverged sweep would trade a *stated* bias for
an unstated one.

## §3 — Whole-shot marching (`solvers/mpm/wholeshot.py`)

**The declared-approach path is untouched.** `march()`'s signature, `_approach`,
`prepare()` and `solve()` all behave exactly as before, because that
constant-velocity approach is what makes an F1 answer comparable to F0's
memoryless one.

`simulate_f1_shot` is additional. It places the head just above the sand at its
delivered attitude and integrates it: the sand wrench decelerates the head, the
next pose follows from the last, and the wrench history comes off one continuous
solve. It returns a `ShotResult`, in the shape of
`solvers/shot.py::simulate_shot`, so an F1 shot enters `bunkershot3d.metrics`
through the same door an F0 shot does.

On one 40 x 16 mm sole at 12 m/s and 20 deg attack, 4 mm grid, 30 mm declared
width:

| | declared approach | whole-shot march |
|---|---|---|
| steps / span | 365 / 3.90 ms | 1123 / 12.0 ms (windowed) |
| peak load | 1.06 kN, 1.04 ms after first contact | 534 N, 0.13 ms after first contact |
| head speed | prescribed 12 m/s throughout | 12 → 9.40 m/s, 0.984 N.s impulse |
| sole depth reached | the queried 8 mm | 25.6 mm, over 122 mm of travel |
| divot | 9.08 mm | 26 mm |
| force at 8 mm sole depth | 42.5 N (windowed, at the queried pose) | **17.3 N** |

The 2.5x gap at the *same* sole depth is the whole point: the marched head has
slowed and is ploughing through its own divot, while the declared path drives a
constant-velocity body into an undisturbed bed.

Cost is as ADR-0033 estimated: **23.75 ms/step at 2,600 particles**, so a whole
shot is tens of seconds. A study tier, not the design sweep.

Two things are **declared allowances**, not predictions, because how far a head
travels while submerged is one of the things the march is run to find out:
`travel_allowance_m` and `ejecta_headroom_m`. A head that reaches the end of the
bed raises and names the setting, because past the end it sweeps empty grid and
reports a zero wrench — which reads exactly like an exit.

The strike is bracketed on the sole's own geometry rather than on the contact
set. A swept-node contact fires while the body is still a step's travel above the
surface, so a march that opened its window on "any contact" would close it again
on the same step and report a strike that never happened.

## Structure

`solver.py` was 1077 lines of a 1200-line budget, so this extracts rather than
inflates:

- `contact.py` (279 lines) — `contact_order`, `apply_body_contacts`,
  `push_out_bodies`, `ledger_from_impulses`, `BodyContact`;
- `step.py` (337 lines) — `StepContext`, `advance_step`, `StepDiagnostics` and
  the Cauchy-stress helpers, so a step-by-step caller (which a whole-shot march
  has to be) is a *caller* of the solver rather than a friend of it;
- `ball.py`, `wholeshot.py` — the new behaviour.

`RigidSection.project_grid_velocity` was 104 lines against the 100-line
changed-file budget — already over on main, and only in the checked set now that
`body.py` is touched. It splits into `coulomb_cone_projection` (a free function,
so the branch that matters can be tested directly: the sliding case *subtracts*
the friction bound from the tangential speed rather than replacing it, and
getting that backwards would brake a frictionless club without failing any test
that only exercised sticking) and `RigidSection._active_nodes`.

`solver.py` is now 956 lines. `march()` keeps its exact signature;
`march_bodies()` is the sequence form; `run(state, extra_bodies=...)` reaches it.
`_require_courant` now checks **every** body — the ball is slow at the start of a
shot and fast once the club has reached it, and a condition that only ever looked
at the club would go quiet exactly when it started to matter.

`fields/capture.py::_shifted` now re-times through `dataclasses.replace` instead
of a field-by-field rebuild, which would have silently dropped the new per-body
ledger.

## Verification

- **117 new tests** (40 ball, 25 multi-body, 45 whole-shot, 7 friction cone).
- **871 pass** across `tests/bunkershot3d/solvers/`, `fields/`, `metrics/` and
  `test_public_api_8608.py`; the headless-import guard
  (`test_package_import_stays_headless_9145.py`) also passes.
- `ruff check` / `ruff format` clean; CI-mode mypy
  (`--follow-imports=skip --ignore-missing-imports`) clean on every touched file,
  including a pre-existing `var-annotated` error in `body.py` that this fixes.
- `check_file_size_budget.py`, `check_architecture_budget.py`,
  `check_version_consistency.py`, `check_todo_discipline.py` and
  `check_suppression_ratchet.py` all pass; the suppression ratchet *improves*
  (bare `type: ignore` 60 → 59, bare `noqa` 3 → 2).
- No new dependencies: numpy only.
- Every guard is a `raise`, never an `assert`.

SPEC.md entry at **1.0.624** (next free version above main's 1.0.623).

Refs #8733 (sections 1-3; sections 4 and 6 remain open), #8712, epic #8699,
ADR-0033.

Opened as a **draft** so nothing auto-merges it.


dieterolson:
CI note: `unit-test-gate` is red on two tests in `tests/tools/sidekick_tool/test_embed_adapter.py` (`test_package_registers_adapter_on_import`, `test_package_registration_is_idempotent`). That is issue #9168, not this change:

- nothing here touches `src/tools/sidekick`, the launcher-embed registry, or anything they import — this branch is `src/bunkershot3d/` only;
- the same two tests fail identically on the unrelated sibling branch `feat/f1-verification-remainder` (run 33236820836, a different test total of 12317 against this branch's 12398), so it is not an artefact of this branch's collection;
- they pass in isolation locally (16/16 in that file);
- PR #9175 is the fix for #9168.

Everything else in the aggregate is green: `code-quality`, `security-scans`, `repo-structure-gates`, `tests`, `rust-wheel-parity`. `quality-gate` fails only because it aggregates `unit-test-gate`.

The first `repo-structure-gates` failure on this branch was real and is fixed in c9a9e23: `RigidSection.project_grid_velocity` was 104 lines against the 100-line changed-file budget — already over on main, and only in the checked set once `body.py` was touched. Worth knowing for anyone else who edits that file: the gate reads the **working tree** but selects files from `origin/main...HEAD`, so running it before committing does not see the change you are about to make.

PR #9234, opened 2026-08-29
fix: stop the tests(3.12) hang inside subject_scaled_closed_contact

dieterolson:
## Summary

Fixes the `tests (3.12)` hang on `main` (job [99142314887](https://github.com/D-sorganization/UpstreamDrift/actions/runs/33268339977/job/99142314887)), which currently blocks merges.

- The job's own faulthandler dump shows it stuck inside `least_squares` -> `solve_closed_contact_configuration` -> `_contact_kinematics` -> `_point_jacobians`, called from `tests/research/test_subject_scaled_closed_contact.py::test_small_closed_contact_atlas_reports_every_feasibility_gate`.
- **Investigated and ruled out** the two hypotheses the trace first suggests:
  - *Unbounded iterations*: `max_nfev=1000` is already set explicitly, and a sweep confirms the solver does terminate (never truly infinite) -- just slowly.
  - *Wrong jacobian*: a finite-difference check of the analytic jacobian in `_contact_kinematics`/`_point_jacobians` (the exact frames in the CI trace) matches to relative error ~2e-10 -- it is correct.
  - *Silent non-convergence*: the module already reports `solver_converged`/`feasible` as structured dataclass fields rather than hanging or raising, and a dedicated test (`test_unreachable_grip_span_is_retained_as_adverse_control`) depends on that graceful-report behavior for a deliberately infeasible case.
- **Root cause**: `ClosedContactConfig`'s default `ftol=xtol=gtol=1e-11` is ~7 orders of magnitude tighter than the `closure_tolerance_m=5e-4` gate that actually determines feasibility. Contact closes by iteration ~10-20; the remaining ~200-500 iterations (measured `nfev` range: 203-500 across the 12-case grid) are SciPy's trust-region optimizer chasing a near-flat regularization tail (least-norm tie-breaking among the reduced-tree's redundant DOF) with no engineering-relevant improvement (`hand_to_grip_distance_m` moves from ~6.7e-8 m to ~6.6e-11 m, both far inside the 5e-4 m gate). At ~2-3 s/solve x 12 solves, that's ~23-30 s for this one test -- comfortably over budget on a contended CI runner even with a 60 s per-test timeout.
- **Fix**: `test_small_closed_contact_atlas_reports_every_feasibility_gate` -- which only asserts array shapes/metadata, never convergence or feasibility outcomes -- now passes an explicit `ClosedContactConfig(solver_tolerance=1e-6, maximum_function_evaluations=60)`. The module's published default is left untouched: it's hash-pinned by `source_sha256` in the committed evidence bundle (`docs/research/proximal_distal_energy_transfer/data/subject_scaled_closed_contact.{json,npz}`) and consumed by roughly a dozen downstream research scripts, and the two single-solve tests in this file intentionally exercise it at full precision.

## Timing (red/green)

| | whole file (`-p no:cacheprovider`) | the slow test alone |
|---|---|---|
| before (red, unmodified `origin/main`) | 30.66 s, exit 1 | 22.84 s |
| after (green, this branch) | 15.23 s, exit 1* | 6.99 s |

\* Exit 1 in both runs comes from `test_committed_closed_contact_evidence_and_figure_are_current`, which fails identically before and after this change -- a pre-existing `spatial_full_body.py` source-hash drift unrelated to this fix (confirmed: that file is untouched by this PR, and the same hash mismatch reproduces on a clean checkout of `origin/main`).

## Does the solve converge, or report failure honestly?

Both, depending on the case -- and that was already true before this PR:
- For the two single-solve tests (`test_solver_closes_both_contacts_without_moving_the_club`, `test_unreachable_grip_span_is_retained_as_adverse_control`), the module already reports `solver_converged` / `feasible` as explicit boolean fields on `ClosedContactSolution`, computed from `result.success` and the engineering gates (closure/limits/collision/rank). Non-convergence is a **reported outcome**, not a hang -- this is exactly what the task's "bound the solve" guidance asks for, and it was already implemented.
- The atlas-shaped test this PR touches doesn't need scipy's own convergence flag to be true (it only checks shapes/metadata), so speeding it up with a tighter iteration cap doesn't change what it validates.

## Other findings (not addressed here)

- Issue #9204 catalogs the same "`scientific`-marker tests are not deselected from the default lane" pattern across several other files (`test_shoulder_velocity_strategy_search.py`, `test_engine_tiers.py`, etc.). This PR fixes one concrete instance; the systemic lane/budget question is explicitly left to a "deliberate decision" per that issue.
- `tests/research/test_closed_state_forward_bridge.py::test_all_closed_states_map_with_declared_closure_gates` (also under `scripts/research/proximal_distal_energy/`) showed a similar slow-but-not-fatal signature in the same CI job (a non-fatal SIGUSR1 hang-forensics dump at 18:46:20, mid `map_closed_contact_atlas`). It's a different mechanism (unconditionally iterating all 234 committed atlas samples, not a solver-tolerance issue) and out of scope here; flagged separately.

## Test plan

- [x] `pytest tests/research/test_subject_scaled_closed_contact.py -p no:cacheprovider -v --durations=20` -- 4 passed, 1 pre-existing unrelated failure (see above)
- [x] `ruff check` / `ruff format --check` on the touched test file
- [x] `mypy --follow-imports=skip --ignore-missing-imports` on the touched test file
- [x] Finite-difference verification of the analytic jacobian (scratch script, not committed)
- [x] SPEC.md changelog row added (1.0.652) and `npx prettier@3.1.0 --check` passes


dieterolson:
Triage from the fleet PR-iteration cycle — two independent failures here:

1. **`unit-test-gate` (and therefore the required `quality-gate`) is not yours.** It is red on all six open PRs in this repo with one failure out of ~12,900: `test_bootstrap_handles_import_errors` asserting `['swing_objective_lab'] == []`. Root cause and fix in [#9272](https://github.com/D-sorganization/UpstreamDrift/pull/9272) — the test harness intercepted adapter imports by the substring `_embed_adapter`, which two entries in `FALLBACK_ADAPTER_MODULES` do not contain. Nothing for you to do; this PR's gate should clear once #9272 lands.

2. **`tests (3.11)` / `tests (3.12)` is yours, and it is a last-bit float comparison.** `test_committed_closed_contact_evidence_and_figure_are_current` fails with:

   ```
   Mismatched elements: 1 / 6 (16.7%)
   Max absolute difference among violations: 3.46944695e-18
   Max relative difference among violations: 1.57415934e-16
   ```

   The printed ACTUAL and DESIRED arrays are identical to six decimals — one element differs by one ULP. That is an exact-equality assertion over a recomputed float, which is not reproducible across BLAS builds or interpreter minor versions. Left alone deliberately: this is the module your PR exists to fix, so the tolerance is your call, not a mechanical rebase fix.

Issue #8735, opened 2026-08-17
unit-test-gate: 453 pre-existing unit failures, clustered into 17 root causes


dieterolson:
`unit-test-gate` on a clean `origin/main` (0065ec3bb):

```
453 failed, 10531 passed, 136 skipped, 19 errors in 372.28s
```

This is pre-existing debt, not new breakage. `quality-gate` was pinned to
`d-sorg-fleet-docker` and **no `CI Standard` run in the last 30 reached success or
failure at all**, so this job's result was never observed. #8729 fixed the routing; this
is the first honest read of it.

**This check cannot be made green without clearing the backlog below, and it should not
be made green any other way.** Filing rather than patching.

## Root-cause clustering

17 clusters account for ~306 of 453 failures and 18 of 19 errors (~69%). The remaining
~147 failures are a genuine long tail across ~85 files with no shared signature.

| # | Cluster | Tests | Type |
|---|---|---|---|
| C1 | `_prevent_repo_root_io` chdir vs repo-relative `open()` | 41 | test |
| C3 | `hatchling` stub import-order collision | 29 | test |
| C2 | `theme` / `ui` package exports dropped, production still imports them | ~35 | **production** |
| C4 | Sidekick ownership manifest | 23 F + 5 E | **production** |
| C5 | `src.*` alias registration never binds parent attributes | 23 (+9) | **production** |
| C6 | `solver_status` rename escaped its scope into 2 test files | 23 | test |
| C17 | Rust `upstream_physics` kernel absent from the gate venv | 22 | CI config |
| C12 | Guard exception-type drift (`assert` → `ValueError`) | 17 | test |
| C11 | Data-explorer durable store not isolated between tests | 15 | test |
| C7 | `PureWindowsPath` / `WindowsPath` misuse | 14 | test |
| C8 | WebSocket local guard landed without updating WS tests | 13 | test |
| C9 | CORS DbC precondition vs `MagicMock` | 11 | test |
| C10 | `Styles.SIDEBAR_MIN_WIDTH` / `MARGIN_PAGE` referenced, never defined | 10 | **production** |
| C13/C14 | `InertiaResult` / `WorkspaceRegistry` APIs tests were written against but never implemented | 16 | feature |
| C15 | `MagicMock` model passed into real MuJoCo C bindings | 8 | test |
| C16 | Pose Studio cannot construct | 1 F + 7 E | **production** |
[...]

PR #9254, opened 2026-08-30
fix(bunkershot3d): un-mirror the delivery frame so bounce skids instead of digging (#9247)

### LINKED ISSUE #9247: F0 inverts the primary design variable: more marketed bounce digs deeper, not shallower
Bounce exists so the sole skids instead of digging. F0 models the opposite. Reproduced on `main` (`0c064b533`) through the public workbench API, 24 mm sole, −14° attack, firm sand, default swing:

| marketed bounce | sole depth | divot mass | peak force |
| ---: | ---: | ---: | ---: |
| 8° | 19.69 mm | 109.3 g | 1188.0 N |
| 14° | 21.40 mm | 122.3 g | 1241.1 N |
| 20° | 22.99 mm | 135.3 g | 1152.3 N |
| 26° | **24.57 mm** | **149.2 g** | 1135.1 N |

**Monotone across the whole range: +18° of bounce buys 4.9 mm *more* depth and 40 g *more* sand moved.**

This is the primary design variable of a wedge. A tool built to compare sole geometries cannot rank them while the sign of its main lever is inverted, and every bounce-related conclusion the model has produced is suspect until this is resolved.

## Why this is probably a measurement artefact, not a force-model error

Peak force **falls** as bounce rises (1188 → 1135 N) while depth and displaced mass **rise**. Deeper, more sand moved, less peak force is not a self-consistent combination for a resistive medium: the depth-linear term should grow with depth and engaged area. That points at the depth *reference* rather than at the wrench.

Two concrete candidates, both cheap to test:

1. **The sole reference point moves with bounce.** #8701 split `depths_m` into `engaged_depths_m` and a geometric `sole_depths_m` measured "from a named reference point". If that reference is defined on the sole in a way that shifts as bounce changes the sole's curvature, then increasing bounce relocates the origin of the measurement and inflates reported depth without the head actually sitting lower in the bed.

2. **`ShotSettings.start_at_first_contact` interacts with it.** The head is dropped so its lowest element begins exactly at the free surface. As bounce changes, *which* element is lowest changes. If the drop uses one point and the depth report uses another, the two disagree by a bounce-dependent offset — exactly the shape of this error.

A third possibility, which the data argues against but which must be excluded rather than assumed: the delivered pose genuinely places a higher-bounce head deeper, in which case `deliver_wedge`'s effective-bounce trigonometry is wrong.

## What a fix has to establish

- **Which quantity is wrong** — depth reference, delivered pose, or wrench — demonstrated by measurement, not chosen by plausibility. The three hypotheses above make different predictions about `engaged_depths_m` vs `sole_depths_m` vs the pose, so they are separable.
- **A regression test asserting the physical ordering**: at fixed loft, attack and sand, more marketed bounce must not produce a deeper sole or more displaced mass in the non-burying regime. This is the invariant that was missing, which is why an inverted primary lever survived a suite of 2,000+ tests.
- Whether the burying regime genuinely flips the ordering (prior work observed it does, at −10° with a 5° sole burying 75 mm vs a 14.42° sole at 13 mm) — if so, the test must encode *both* regimes and the boundary between them, not just one.

## Provenance

Surfaced by the dig/skid rebuild (#8703, PR #9171), which was the first metric able to see it — the previous discriminant was saturated and returned `MARGINAL` for every design, so it could not have exposed an inverted lever. `DIG_SKID_BOUNCE_ORDERING_REASON` currently travels with every verdict so nobody reads one as a bounce recommendation; that caveat is a stopgap and should be removed by this fix, not left as permanent cover.

Filed now because it was previously recorded only as an inline caveat and a task chip, with no issue tracking it.

dieterolson:
Closes #9247.

The delivery frame was mirrored against the head frame. F0 drove the wedge through the sand **trailing edge first**, down a sole descending at 19.97 deg into the bed, on which more bounce is a steeper ramp. The tool's primary design variable came out with the wrong sign.

## Before / after

24 mm sole, firm sand, 25 m/s, 10 deg open, 6 deg forward lean — the issue's own reproduction.

**At −14 deg of attack** (`main` at `0c064b533`, left; this branch, right). The window is widened to 400 ms after the fix so a buried head still reports a depth and the two regimes share one axis.

| marketed bounce | presentation bounce | depth (main) | mass (main) | peak F (main) | depth (fixed) | mass (fixed) | peak F (fixed) |
| ---: | ---: | ---: | ---: | ---: | ---: | ---: | ---: |
| 8° | −7.53° | 19.69 mm | 109.3 g | 1188.0 N | **71.66 mm** | buried | 398.9 N |
| 14° | −1.55° | 21.40 mm | 122.3 g | 1241.1 N | **57.16 mm** | buried | 497.8 N |
| 20° | +4.45° | 22.99 mm | 135.3 g | 1152.3 N | 17.95 mm | 154.7 g | 598.5 N |
| 26° | +10.45° | **24.57 mm** | **149.2 g** | 1135.1 N | **16.49 mm** | **110.9 g** | 645.4 N |

**At −4 deg of attack**, where every design in the sweep planes:

| marketed bounce | depth | mass | peak force | exit speed |
| ---: | ---: | ---: | ---: | ---: |
| 8° | 7.03 mm | 59.9 g | 225.6 N | 20.74 m/s |
| 14° | 6.89 mm | 56.6 g | 175.2 N | 23.38 m/s |
| 20° | 6.87 mm | 56.6 g | 155.4 N | 23.87 m/s |
| 26° | 6.83 mm | 55.1 g | 158.8 N | 23.74 m/s |

Over a 7 × 7 sweep of bounce (8–26°) against attack (−2 to −14°), depth is monotone the right way in **49 of 49** cells. On `main` the same 49 cells are monotone **inverted**.

The peak force now reads consistently with the depth: a planing sole meets a shorter, harder contact and keeps its speed (16.5 mm, 645 N, exits at 17.5 m/s), while a buried one decelerates over a long slow crawl (71.7 mm, 399 N, exits at 1.5 m/s). On `main` depth and mass rose while peak force fell, which is what said the defect was not in the wrench.

## Which quantity was wrong, and how the others were excluded

The issue listed three hypotheses. Two were excluded by measurement, not by plausibility.

**1. "The sole reference point moves with bounce" — excluded.** Its body-frame station is **25.803 mm at every bounce in the sweep**, to the micrometre. Only its `z` moves, and that is a rigid offset of the body origin, not a change of measurement origin. `engaged_depths_m` also tracked `sole_depths_m` to within 0.1 mm throughout (19.59 vs 19.69, 24.52 vs 24.57), so the two independent depth measures agreed: the head genuinely went deeper.

**2. "`start_at_first_contact` drops on one point and reports another" — excluded.** The mismatch is real (the workbench passes the address-lowest sole element while the head is delivered rotated) but it is a pre-burial of **0.034–0.042 mm which *shrinks* as bounce rises** — two orders of magnitude below the 4.9 mm swing, and in the wrong direction.

**3. The delivered pose — this one, but not as `deliver_wedge`'s trigonometry.** Its angles are right. The frame its *vectors* are expressed in was not:

- `geometry/profile.py` and `geometry/lofting.py` build the mesh with the leading-edge point at the origin and **`+x` rearward**;
- `geometry/delivery.py` was written with **`+x` toward the target**.

Two contradictory frame conventions in one package, each stated only in prose, in a docstring. `delivery.py`'s scalars never depended on the sign of `x`, so they were right; its vectors were backwards, and the workbench — which rebuilt the delivery composition from that module's docstring and took `+x` for the travel direction — inherited both halves.

**Measured on the lofted mesh rather than argued.** The face normal at address is `(-0.516, 0, 0.856)`: it points along `-x`. Delivered loft read off the mesh now tracks `effective_loft_deg` to a **constant 0.920 deg** discretisation offset at every open/lean combination, against errors of **−7.81 to +28.92 deg** before:

| face open | shaft lean | mesh − library (main) | mesh − library (fixed) |
| ---: | ---: | ---: | ---: |
| 0° | 0° | 0.92° | 0.92° |
| 0° | 6° | 12.92° | 0.92° |
| 0° | 14° | 28.92° | 0.92° |
| 10° | 0° | −7.81° | 0.92° |
| 10° | 6° | 4.04° | 0.92° |
| 20° | 8° | −1.11° | 0.92° |

At the shipped 6 deg of forward lean the sand saw a 64.92 deg face while the report said 52.00 deg, because the mirrored pitch *added* loft instead of removing it. And the surface meeting the sand head-on was the **rear of the flange**: the face engaged **0.00 cm²** on entry under `main`, against 17.31 cm² at 14 mm of depth after the fix.

## What changed

`delivered_rotation` is exactly `M R M` with `M = diag(-1, 1, 1)` — a mirror in `x`, which reverses rotations about the shaft axis and about `y` alike. **Every published angle is unchanged**, verified bit-identical over 576 loft/bounce/lie/open/lean combinations. Only vectors moved.

- `geometry/delivery.py` restated in the head frame; `delivered_rotation`, `entry_velocity_m_s` and `TRAVEL_AXIS_BODY` are now public, so the composition and the travel axis are stated **once** rather than reimplemented from a docstring across a package boundary.
- `geometry/lofting.py` gains `require_travel_frame`, a precondition on every lofted head: the sole must descend rearward from the leading edge. `raise`, not `assert` — `python -O` strips assertions and every delivered pose in the package rests on this. (Note the obvious version of this check does not work: an `x` mirror maps extremes to extremes, so comparing the mesh's own extents passes either way round. It is the leading edge's forward-referenced station that breaks the symmetry.)
- `bunker_shot_gui/bridge.py` delegates to the library instead of rebuilding the composition; `model.py`'s scene travel axis follows `TRAVEL_AXIS_BODY`.

## The invariant that was missing

`tests/bunkershot3d/solvers/test_bounce_ordering_9247.py`, 19 tests. An inverted primary lever survived 2,000-odd tests because every one of them pinned a shape, a conservation identity or a scalar against itself, and none said which way round the answer should come out.

Both regimes and the boundary, because a test pinning only the shallow case would be half a guard:

- **non-burying**: more bounce is never deeper and never moves more sand, at −2/−4/−6 deg, square *and* at the shipped 10 deg open / 6 deg lean (a square delivery leaves the rotation at identity and cannot see a mirrored rotation at all);
- **burying**: the ordering does not reverse across the boundary;
- **the boundary** is bracketed rather than pinned — one steep sweep must contain both a burying and a non-burying design.

Each half of the mirror is caught independently: mirroring the delivered rotation alone fails 5 tests, mirroring the travel direction alone fails 9, mirroring both fails all 17 at the new lofting precondition. The frame tests measure the pose against the **published first-order rates**, not against `delivered_rotation` — a test that measures a rotation against itself passes in either frame, which is how a mirror survives. (An earlier draft of this test did exactly that and passed under a mirrored rotation.)

## Two findings a reviewer should not miss

**The burying-regime "reversal" is not real.** Prior work read the ordering as flipping where the head buries. After the fix more bounce is shallower in *both* regimes — they differ in magnitude, not in sign. The apparent flip was the *shallow* regime being inverted, not the deep one. The boundary sits near zero presentation bounce (delivered bounce plus attack angle) and drifts a couple of degrees positive as the attack steepens.

**The corrected model buries three of six shipped presets at the workbench's default delivery** (−8 deg attack, 10 deg open, 6 deg forward lean), because their presentation bounce is negative there: `tour_shaved_heel_lob` (−4.5°), `sm9_58_m` (−1.5°), `acushnet_example_1` (−0.5°). That is the model's honest output rather than a new defect — and it is bounded by the standing flat-half-space caveat, which has no crater and no divot memory, so deep digs are over-predicted. It does make the default delivery worth revisiting: 6 deg of *forward* shaft lean is a full-shot delivery, not a splash shot, and under the mirrored frame it was silently *adding* 6 deg of bounce instead of removing it. Not changed here.

## Stopgap removed

`DIG_SKID_BOUNCE_ORDERING_REASON` rode on every dig/skid verdict to warn readers off the bounce ordering. It covered this bug; leaving it after the fix would convert a temporary warning into permanent noise. Removed, and the test that pinned it in place is inverted rather than deleted so the removal is owned by an assertion.

**Merge note:** `fix/divot-mass-undercount` is also editing `metrics/divot.py` and its hunks touch the `__all__` block, so expect a one-line conflict there. The two changes are independent; keep both.

## Verification

- `tests/bunkershot3d/geometry` + `tests/bunkershot3d/solvers`: 994 passed, 1 skipped (optional `trimesh`).
- `tests/bunkershot3d/metrics`: 162 passed.
- `tests/tools/bunker_shot_gui` + `tests/bunkershot3d/study` + `tests/bunkershot3d/vandv`: all passed after the one change below.
- `ruff check` / `ruff format` clean; `mypy` clean on the five changed source files (the 3 remaining errors are pre-existing, in untouched `src/shared/python/`).

One existing test changed. `test_both_tiers_oppose_the_head` asserted that the vertical sand force is upward at every cross-tier probe — the same claim as "the sand opposes the head" only while the sole *planes*. Once the presentation angle goes negative the face submerges and the sand on it drives the head down; that is what digging is. It could only ever pass because the mirrored frame had the head striking with its back. Restated as the law it meant, `F · v ≤ 0`, with equality allowed only where the solver reports no force at all. Measured at the default delivery: `F · v` is −61, −6822, −0.0 for F0 and −2719, −1085, −3.4 for F1 across the three probes.


dieterolson:
Resolved the `SPEC.md` conflict and fixed one of the three failing gates. Triage of the rest, so the remaining work is clear — I did **not** guess at the physics.

**Fixed: `repo-structure-gates`.** The changelog had `1.0.658` on two rows: this branch's delivery-frame entry and `main`'s divot-mass entry, which claimed 658 while this branch was open. Renumbered this branch's row to **1.0.665** (main is at 1.0.661; 662/663 are held by #9234/#9252, 664 by #9253) and moved the identity `Spec Version` field with it. `check_spec_changelog_duplicates.py` passes. That also clears `test_check_spec_changelog_duplicates.py::TestTheRealSpecIsClean`, one of the 19 unit-test-gate failures.

**Not fixed, needs you — 10 × `TypeError` in `tests/bunkershot3d/solvers/test_bounce_ordering_9247.py`.** `main` made `friction_angle_deg` a required keyword on `divot_metrics()` (SPEC 1.0.658), and `_shoot()` in this branch's new test file still calls it with only `width_m` and `bulk_density_kg_m3`. The fix is mechanical *in shape* — the module already declares its bed as `playing_condition(PlayingCondition.FIRM)` in the `sand` fixture, and `main`'s own tests thread `friction_angle_deg=sand.friction_angle_deg` — but `_shoot()` has no `sand` parameter today, so wiring it means touching the helper and all seven call sites inside your test file. I left that to you rather than reshaping an actively-developed test module. Note the API change is deliberately default-less ("the wall angle is the bed's and a default would be an invented divot shape"), so picking a literal here would be the wrong fix.

**Not this PR — 5 × assertion failures in `test_workbench_model.py` / `test_shot_scene.py`.** `main` is red here independently; `tests (3.11)`/`tests (3.12)` and `quality-gate` have been failing on `main` since ~2026-08-29T17:37Z, tracked as #9204 (suite exceeds its job budget) and #8735 (pre-existing unit failures). Running `test_workbench_model.py` against `origin/main` alone reproduces 11 failures (`test_carry_is_reported_for_a_greenside_shot` → `assert nominal_shot.carry_m is not None`, plus the playability-window and comparison classes), so those are inherited, not introduced here. Worth confirming against #9204 before spending time on them from this branch.

`test_sole_load_render.py::test_every_frame_of_the_shot_can_be_drawn` died as `worker gw0 crashed` — the known pytest-qt/xdist crash pattern, not an assertion.

PR #9374, opened 2026-09-02
fix(scripts): repair validate_suite.py for the current src/ layout

### LINKED ISSUE #8834: scripts/validate_suite.py validates a pre-migration layout and can never pass
Found during the 2026-08-21 adversarial integration review (PR #8815).

`scripts/validate_suite.py:86,94` does `from launchers.unified_launcher import UnifiedLauncher` and `from launchers.golf_suite_launcher import main`. There is no top-level `launchers/` package, and the script only inserts the project root on `sys.path` (lines 22-24), not `src/`, so both imports raise `ImportError` and `validate_launchers()` returns `False` unconditionally. Its `expected_dirs` list (lines 41-54) is resolved against `get_src_root()` and includes `docs` and `output`, neither of which exists under `src/`.

The script is a leftover from a directory reorganization that was never completed — and it is one of the only remaining "consumers" keeping the deprecated launcher shells alive (see #8831).

**Suggested fix:** delete it, or rewrite it against the current layout and wire it into CI so it can't rot again.

Severity: medium.

dieterolson:
## Summary
`scripts/validate_suite.py` validated a pre-migration directory layout and could never pass -- only 1/6 validators succeeded when run against current `main`.

- `validate_launchers()` never put `src/` on `sys.path`, so the launcher import chain's bare `import bunkershot3d` (used transitively by `simulation_backends/wrench_extractor.py`) failed with `ModuleNotFoundError`, even though the launcher code itself is fine.
- `validate_shared_components()` imported `shared.python...` instead of `src.shared.python...` (there is no top-level `shared` package), and separately imported `SUITE_ROOT` from `src.shared.python.core.constants` -- a *different*, more narrowly scoped path (`src/shared`) that shadowed the script's own module-level `SUITE_ROOT` (`src/`) for the rest of the function, so the MATLAB file check below it looked in the wrong place.
- `validate_git_repository()` and `validate_configuration_files()` resolved `.git`, `.gitignore`, `README.md`, and `docs/` against `get_src_root()` (`src/`) instead of the project root, where they actually live.
- `validate_directory_structure()`'s `expected_dirs` included `docs` and `output`, resolved against `src/` -- neither exists there; both exist at the project root.

## Fix
Straightforward import-path/root fixes -- no directory reorganization, no behavior change to what's being validated, just pointing each check at the right root/module path.

Fixes #8834

## Test plan
- [x] New `tests/scripts/test_validate_suite.py` (7 tests, one per validator plus the end-to-end run) -- 5/7 failed before the fix, all 7 pass after
- [x] `python scripts/validate_suite.py` run directly: 1/6 -> 6/6 validations passed
- [x] `ruff check` / `ruff format --check` on changed files


dieterolson:
The `unit-test-gate` failure here (`module 'src' has no attribute '_install_parent_shared_aliases'`) doesn't reproduce locally — 7/7 tests in `tests/scripts/test_validate_suite.py` pass cleanly in a fresh clone. Filed #9387 with evidence this is a recurring xdist worker-corruption issue (same class as #9099, different victim) affecting this PR and #9384 simultaneously — not a defect in this PR's own diff. Leaving auto-merge armed; this should clear once #9387 is root-caused.

Issue #9352, opened 2026-09-01
ADR-0047 H3: BallFlight web page imports trajectory records


dieterolson:
Part of accepted ADR-0047; blocked on H1. The /ball-flight page + api/routes/ball_flight.py accept an uploaded/selected record for overlay, provenance shown.

dieterolson:
```yaml
agent: claude
session: adr0047-h3-ballflight-import
claimed_at: 2026-09-02T06:02:54.546634+00:00
expires_at: 2026-09-02T08:02:54.546634+00:00
intent: implement
```

PR #9384, opened 2026-09-02
feat(shot-tracer, ADR-0047 H2): import cross-family trajectory records

### LINKED ISSUE #9351: ADR-0047 H2: Shot Tracer imports trajectory records
Part of accepted ADR-0047; blocked on H1. Shot Tracer's multi-model comparison accepts imported `ball_flight_trajectory/1` records alongside its native models — a Tools swing_sim flight can finally sit on the same axes as Waterloo-Penner, each labeled by provenance.

dieterolson:
## Summary

- Shot Tracer gains an **"Import Trajectory Record…"** action that reads a `swing_sim.ball_flight_trajectory/1` JSON record through the **vendored** Tools `flight_interchange` reader (resolved via the canonical `tools_repo_path.resolve_tools_repo` facade — the same posture `external_tools_adapter.py` uses) and plots it alongside the native Waterloo/Penner-family curves.
- Every imported curve is labeled `"<model_family> / <model_name>"` in a new **Imported Trajectories** list widget — never plotted unlabeled.
- Frame handling is explicit and closed: only the wire's `flight_xfwd_yleft_zup` frame (Shot Tracer's own native plot frame) has a converter. `app_xtarget_yup_zright` is a legal wire frame but has no converter yet, so it is refused with a named reason rather than silently mis-plotted.
- Every refusal — unsupported frame, or any wire violation surfaced by the vendored reader (unknown/missing fields, malformed provenance, non-monotone samples, an unavailable vendor checkout) — reaches the user as a dialog carrying the *specific* reason. Nothing is dropped silently.
- New module `src/launchers/_shot_tracer_trajectory_import.py` is the Qt-free import/validation/frame-conversion logic; `_shot_tracer_gui.py` wires it to a button, the imported list, and the GL plot.
- `simulation.shot_tracer` in the feature-parity registry moves from `parity` to `gap` (issue #9352 / ADR-0047 H3), since the PyQt tile now has cross-family import that the web BallFlight page does not yet.

Depends on the `vendor/ud-tools` pin advancing past Tools#4888 (`swing_sim.flight_interchange`) via #9363.

Closes #9351

## Test plan

Headless (`QT_QPA_PLATFORM=offscreen`), against the *real* vendored reader — no mocking of `ball_flight_trajectory_from_json`:

- [x] `tests/launchers/test_shot_tracer_trajectory_import.py` (9 tests): a UD-family record (via `flight_trajectory_export`, H1's own export path) and a hand-authored Tools-family record both import with correct labels/positions; an app-frame record and an unknown-field record are refused with the named reason; malformed JSON, a missing file, a non-`Path` argument, an unavailable vendor checkout, and a malformed provenance section are each refused explicitly.
- [x] `tests/launchers/test_shot_tracer_import_gui.py` (10 tests): the import button, the imported list, provenance labeling, refusal dialogs (including the exact reason text), Clear resetting imported state, and GL-plotted imported curves alongside native results.
- [x] Existing `tests/launchers/test_shot_tracer.py` (15) and `tests/unit/test_shot_tracer.py` (7) unaffected — 41/41 pass together.
- [x] `tests/config/feature_parity/` (39 tests) — registry/matrix freshness gates pass.
- [x] `ruff check` / `ruff format --check` clean on all changed files.
- [x] `mypy` clean on both new/changed files (pre-existing unrelated errors in a transitively-imported module and one pre-existing line in `main()` confirmed unchanged).
- [x] `scripts/ci/check_error_handling_ratchet.py` — improved, no regressions.
- [x] File size budget: largest changed file is 656 lines (`_shot_tracer_gui.py`), well under 1200.


dieterolson:
Update: the vendor-skip fix worked as intended — `test_shot_tracer_trajectory_import.py`'s 6 vendor-dependent tests now cleanly skip in `unit-test-gate` (job run 33627263609) with the expected reason, instead of failing on an unrelated message.

3 failures remain in `test_shot_tracer_import_gui.py`, but they don't reproduce locally: running that file alone, or together with `test_shot_tracer_trajectory_import.py`, passes 100% of the time (19/19) with the real vendor present. All 3 failing tests fully mock `import_trajectory_record` via `@patch("src.launchers._shot_tracer_gui.import_trajectory_record")` and never touch the vendor. This points to cross-test pollution specific to the full `unit-test-gate` xdist run (13,000+ tests) rather than anything in this PR's own tests — the same class of test-order-dependent failure already tracked in #9374 (`sys.modules` shadowing between UD's own `shared.python.*` and the vendored copy inserted onto `sys.path` by `_load_vendored_reader()`). Leaving auto-merge armed; this should go green once #9374 is fixed, and I don't want to mask it with a speculative change to shared import-handling code that every other vendored-Tools consumer in this repo also relies on.

dieterolson:
Filed #9387 tracking the worker-corruption pattern with full evidence from this PR's run plus #9374's.

Issue #9348, opened 2026-09-01
ADR-0046 G1: port-plan inventory for the 30 UD launch-monitor modules


dieterolson:
Part of accepted ADR-0046. A reviewed table classifying every module in `src/shared/python/launch_monitor/` as port-up (UD-only capability -> canonical Tools layer), already-home (Tools equivalent exists; G0 gate covers the pair), or UD-app-local (workbench glue that stays). Output is a doc PR, not code — the port itself happens per-module afterward, tests travelling with each port.

dieterolson:
G0 (#9354) is in with 28 gates and **14 pinned divergences (D1–D14)** — the port plan must start from that evidence, not from module names. Three items that shape G1 directly:

1. **D10 — the estimator disagreement is user-visible.** Identical per-player slopes (delta 0.0) feed two pooled estimators: UD's cluster-robust player-FE OLS says no detectable trend (CI crosses zero, p=0.21); Tools' DerSimonian–Laird random effects says improving with 98% probability (CI excludes zero). The canonical layer cannot ship both as 'the trend' — G1 must classify this as a **named-method pair** (both preserved, results labeled, like the roll models in ADR-0045) or pick one with statistical justification.

2. **Error-handling philosophies differ**: one malformed row in 161 destroys Tools' whole SG result in 3 of 4 modes (and is silently dropped in the 4th, no exclusion record), while UD excludes-and-audits with the mean unchanged. UD's exclude-and-audit posture should be the canonical one — audit trails are strictly more information.

3. **Intra-UD estimand conflict**: UD's SG longitudinal fit is shot-level while UD's own longitudinal.py aggregates to session cells first and warns against pseudo-replication. The canonical layer has to pick one estimand; G1 should name it.

Also from G0's capability matrix: Tools cannot compute SG uncertainty/grouped SG/dispersion ellipse; UD cannot compute lateral sd/RMS, slope uncertainty, heterogeneity (τ²/Q/I²). The union is the canonical target — nothing shrinks.

dieterolson:
Lease: working G1 port-plan inventory on branch `docs/adr0046-g1-port-plan` (worktree `UpstreamDrift-worktrees/lm-port-plan`). Doc-only PR: `docs/adr/0046-g1-port-plan.md`.

dieterolson:
G1 plan is up in #9358 (auto-merge armed, squash).

[...]

PR #9425, opened 2026-09-03
refactor(launch-monitor): Stage 2 wave 1 — six modules retired onto the canonical layer (ADR-0046 G2)

dieterolson:
ADR-0046 Stage 2, wave 1. #9420 cleared the blocker that stopped this; the six modules ADR-0048's port order lists as P1-P6 — every one a tier-0 leaf with no intra-package dependency — are now retired onto the canonical layer vendored from Tools.

The re-point is the mechanical rewrite the port order always described:

```
src.tools.launch_monitor_model.X  ->  shared.python.launch_monitor.X
```

and the **845 lines** of UpstreamDrift implementation behind `dispersion`, `multivariate`, `trends`, `comparison`, `schema`, and `treatment` are deleted. UpstreamDrift runs Tools' code for them now.

## Per-module outcome

| Module | LOC retired | Consumers re-pointed | Retired? |
| --- | ---: | --- | --- |
| `comparison` | 147 | façade (`gui.py` via `compare_monitors`) | yes |
| `dispersion` | 70 | façade (`gui.py`), `test_dispersion_drift.py` direct | yes |
| `multivariate` | 108 | façade (`gui.py` via `compute_pca`/`compute_vif`) | yes |
| `schema` | 195 | façade (`gui.py`, `widgets.py`), **9 intra-package importers**, `test_corpus_drift.py` direct | yes |
| `treatment` | 215 | façade (`gui.py`) | yes |
| `trends` | 110 | façade (`gui.py` via `analyze_trend`) + the `TemporalTrendResult` rename | yes |

Consumers were resolved by AST, not grep — the package `__init__.py` is a re-export façade, so a text grep attributes every consumer to the façade rather than to the module that owns the symbol. That is also why the diff is small: the façade is the seam ADR-0048 identified, and it carries every workbench and API consumer of all six modules (`gui.py`, `widgets.py`, `flexible_analysis_widget.py`, the analytics routes) with no edit to any of them.

`schema` was the only wave-1 module with intra-package consumers: `contract_v2`, `corpus`, `flexible_analysis`, `importer`, `modeling`, `player_covariation`, `profiles`, `project`, `relationships`. All nine now import `shared.python.launch_monitor.schema`.

## `TrendResult` → `TemporalTrendResult`, no alias

ADR-0048 P3 renamed this in the canonical layer deliberately, because `rate_of_closure` exports `TrendResult` for a *different estimand* — a cumulative session-ordinal mean, not a per-day robust slope. An alias here would re-create exactly the collision the rename exists to prevent, so the façade's `__all__` moves to the new name and `test_trends_exposes_the_renamed_result_without_a_back_compat_alias` asserts the old name is absent on both the canonical module and the façade.

## What replaced the parity pins

`test_canonical_layer_parity.py`'s six twin-identity assertions are obsolete by construction: there is no second copy left to compare, and comparing a deleted file to itself is not a gate. They are replaced by `test_wave_1_module_is_retired_and_served_by_the_canonical_layer`, which asserts **both** halves of the claim — the UpstreamDrift file is gone, *and* the canonical module imports and resolves inside `vendor/ud-tools`. Deleting the file without the import resolving would break the workbench; the import resolving while a UpstreamDrift copy quietly returned would mean the retirement never happened. The provenance probe on `shared.python.launch_monitor` is kept unchanged — it is what every remaining wave depends on.

## The canonical import does not resolve outside the test session

The `companion-workflows` gate caught the consequence that no in-session test could, and it is larger than one workflow: `shared.python.launch_monitor` only resolves where `vendor/ud-tools/src` is reachable. The pytest session gets that from `pyproject.toml`'s `pythonpath` and from `tests/conftest.py`; an installed wheel gets it from `build_hooks.py`, which force-includes the canonical `shared` package at the top level. **A process started from a source checkout gets it from neither** — so after wave 1 the façade could not import at all in a running API server, in the launcher, or in the companion workflow.

`_ensure_canonical_layer_importable` in the façade fixes that at the one place every consumer passes through, and is a no-op wherever the vendored source is already reachable. Two details are worth review attention:

* **Putting the directory on `sys.path` is not sufficient.** `shared` is a regular package, so its `__path__` is fixed at first import, and the probe binds `shared.python` to UpstreamDrift's own copy on its way to discovering that `launch_monitor` is not there. The vendored directory is therefore *appended* to that `__path__` — the same thing `tests/conftest.py` does, and the reason the canonical import resolves under pytest today. This is **not** ADR-0048's rejected Option 3: that option kept the old import strings and let deleted modules fall through silently, whereas here the import strings are the canonical ones and UpstreamDrift has no `launch_monitor` under `shared.python` to be ambiguous with. Appending rather than prepending keeps UpstreamDrift's own shared modules first.
* **The vendored files win over the gitlink validation.** `unit-test-gate` has the pinned Tools files present but not as a real submodule — `fetch-pinned-tools` materialises them with `actions/checkout` — and `require_tools_repo`'s git-level validation rejects that. `src/api/routes/_ball_flight_trajectory_import.py` documents exactly this trap; the bootstrap follows the same shape. The validating facade is consulted only when there is nothing on disk to find, which is the part that genuinely needs a precedence order (`TOOLS_REPO_PATH`, sibling discovery).

`test_facade_imports_without_the_pytest_path_wiring` pins it directly: a subprocess with only `.`, `src`, and `src/shared/python` on `PYTHONPATH` imports the façade and lands in the vendored tree. The `companion-workflows` job also gains `fetch-pinned-tools`, like every other job that needs the vendored stack, since the launch-monitor workflow now genuinely depends on the canonical layer.

## Verification

Every module's UpstreamDrift unit tests and the consuming tab tests were run against the canonical import.

| Suite | Result |
| --- | --- |
| `tests/unit/launch_monitor/` | 134 passed, 4 skipped |
| `tests/ui/tools/launch_monitor/` (9-tab workbench) | 22 passed |
| `tests/api/` launch-monitor routes (4 files) | 37 passed |
| `tests/integration/launch_monitor_drift/` | **71 passed — untouched** |
| `tests/companion/` | 57 passed |
| `tests/unit/repo_hygiene/test_no_shadow_of_tools_shared.py`, `tests/config/test_registry_exclusions.py` | 10 passed |
| `ruff check` / `ruff format --check` | clean |
| repo-structure gates, `check_architecture_budget --base-ref origin/main`, changed-file mypy | clean |

The drift gates stay at 71 — their count since #9419 — and are untouched. Re-pointing the UpstreamDrift side of the dispersion gate leaves it measuring the same two programs, because that gate compares against `rate_of_closure.launch_monitor_performance`, which is a different implementation from the canonical layer, not the same one.

Provenance was confirmed directly rather than assumed: `shared.python.launch_monitor.dispersion` resolves to `vendor/ud-tools/src/shared/python/launch_monitor/dispersion.py`, and `.trends` exports `TemporalTrendResult`.

Also removes the `comparison.py` architecture-budget exception #9420 added, rather than renewing it, since the module it covered no longer exists. No ADR edit: ADR-0048 is 55 bytes under the 50KB documentation budget, and this PR executes its port order rather than amending it.

Refs #9348. Next: P7 (`relationships`) is blocked on the D15 follow-up; P10 onward is blocked on G1-D4.


dieterolson:
### `unit-test-gate` is red on `origin/main`, not on this branch

`companion-workflows` is green after the last two commits. The remaining red — `unit-test-gate`, and the `quality-gate` aggregate that mirrors it — reproduces on `origin/main` with none of this PR's changes present, and none of the 13 failing tests touches the launch-monitor package.

Measured by checking out three revisions in the same worktree and running one selection that provokes the pollution — the two directories #9404 changed plus the tests that fail:

```
tests/unit/c3d_viewer/  tests/unit/engines/simscape/three_d_gui/
tests/unit/launchers/test_wsl_probe.py  tests/unit/utils/test_path_validation.py
tests/unit/launcher/test_sidekick_startup_coordination.py
tests/unit/launchers/test_integrations_health_window.py
tests/unit/launchers/test_docker_manager_timeout.py
```

| Revision | FAILED |
| --- | ---: |
| `5ae322996` — #9420, the commit before #9404 | 4 |
| `27b6eeadb` — `origin/main`, i.e. #9404 merged | 15 |
| `32e4eccd5` — this branch | 17 (same set; local Windows adds two) |

Every one of them passes in isolation, so it is cross-test pollution, and it appears at the revision that changed `tests/unit/c3d_viewer/ui/conftest.py` and `tests/unit/engines/simscape/three_d_gui/conftest.py` (+83/+86 lines) to scope the `src` module pivot. The signature fits: `test_path_validation.py` fails with `ImportError: module src.api.utils.path_validation not in sys.modules`, and two `test_integrations_health_window.py` cases that simulate "the shared widget is unavailable" now receive a real `shared.python.ai.mcp.widgets...` object.

CI's own numbers agree: the previous run of this branch was `1 failed, 13092 passed` (that one failure was this PR's new subprocess test, since fixed). After merging `origin/main` it is `13 failed, 13080 passed` — same total, twelve flipped, none of them launch-monitor.

This PR's own suites are green: launch-monitor unit 141, workbench UI 22, launch-monitor API routes 37, drift gates 71 unchanged, companion workflows 14/14. Auto-merge stays armed; it should clear once main is repaired.

dieterolson:
unit-test-gate is green: #9439 fixed the residual pollution from #9404 (restored only the bare `sys.modules["src"]` key on exit, not the whole evicted `src.*` namespace + parent-attribute links — the actual cause of the 15 launcher/Qt/path-validation failures above) and merged to main; this branch was updated onto it, the SPEC.md version-number collision (both branches independently claimed 1.0.711) was resolved by renumbering this PR's row to 1.0.712, and unit-test-gate/quality-gate came back green. Merged.

PR #9441, opened 2026-09-03
fix(tests): fully restore module/path state when exiting the src pivot

dieterolson:
## Summary

Unblocks the red `unit-test-gate` on `origin/main` (and with it the required `quality-gate` aggregate blocking every open PR) by making the C3D-viewer / Simscape `src` module pivot from #9404 genuinely scoped.

#9404 wrapped the pivot in `_enter_pivot()`/`_exit_pivot()`, but the exit only restored the single `sys.modules["src"]` key. `_pivot_sys_path()` mutates much more than that:

- it **evicts every non-`src.shared` `src.*` entry from `sys.modules`**, and
- it **appends `<repo>/src` to `sys.path`** (making the top-level `shared.*` namespace importable).

Both leaked past teardown, so any test that ran after a pivot directory in the same worker saw:

- evicted `src.<sub>` modules — breaking `importlib.reload(...)` (`test_path_validation`: "module src.api.utils.path_validation not in sys.modules") and string-target `monkeypatch.setattr(...)`, which re-imports a *fresh* module copy and patches an object the already-collected test module no longer uses (`test_integrations_health_window`, `test_docker_manager_timeout`, `test_wsl_probe`, `test_sidekick_startup_coordination`, ...);
- a spuriously importable top-level `shared.*` namespace, so tests simulating "the shared widget is unavailable" received the real `IntegrationsHealthDashboardWidget`.

## Fix

The pivot lifecycle in both conftests now, on the **outermost enter**, snapshots the `src`/`shared` namespaces of `sys.modules` and records the `sys.path` entries the pivot adds; on the **outermost exit** it deletes whatever the pivot left in those namespaces, reinstates the snapshot verbatim (preserving module identity), and removes the added path entries. The reentrant depth counter and the three directory-scoped hooks are unchanged, so the #9404 guard `tests/unit/repo_hygiene/test_no_permanent_src_module_shadow.py` passes as written — the no-permanent-shadow invariant is not weakened.

## Verification (local, repo venv, worktree at 27b6eeadb + this fix)

| Check | stock 27b6eeadb | with fix |
| --- | --- | --- |
| Repro selection (c3d_viewer + simscape three_d_gui + 5 victim files, `-q -p no:randomly`) | 15 failed | **4 failed** (= pre-#9404 baseline measured at 5ae322996) |
| `test_no_permanent_src_module_shadow.py` | — | **4 passed** |
| Full `pytest tests/unit/ -n auto` (unique FAILED/ERROR ids) | 848 | **763** |

The local full-gate numbers carry pre-existing environment noise CI doesn't have (missing optional deps like `imageio`/`cv2`/`trimesh`, blocked model downloads). The only 2 ids failing with the fix but not on stock main — a wall-clock performance assertion and a facade-delegation test — both pass in isolation (xdist/load flakes).

SPEC.md change log: 1.0.711.


dieterolson:
Superseded by #9439, which merged first and fixes the same #9404 pollution (full snapshot/restore of the src sys.modules namespace + the pivot's sys.path additions, factored into tests/helpers/engine_src_pivot.EngineSrcPivot and pinned by the strengthened no-permanent-shadow guard). Independently verified #9439's approach: the repro selection returns to the pre-#9404 baseline, and the top-level shared.* sys.modules entries it leaves alone are session-baseline parent-shared aliases, not pivot leakage — probed identical with no pivot directory involved. Nothing from this PR remains unaddressed.

PR #9452, opened 2026-09-03
fix(ci): run manufactured-solution byte-determinism test via rolling profile outside the governed authority stack

### LINKED ISSUE #9451: CI Optional Stack red on main: articulated manufactured-solution test enforces authority patch in non-governed lane
## Defect
`optional-stack-check` is red on main (run 33750885075, 2026-09-03 11:39):

\`\`\`
FAILED tests/research/test_articulated_manufactured_solution.py::test_manufactured_solution_record_is_byte_deterministic
RuntimeError: authority requires exact Python patch 3.11.15
\`\`\`

## Root cause
- \`write_record()\` defaults to \`profile="authority"\`, and \`validate_authority_environment()\` requires exact CPython 3.11.15 + hash-locked dependency versions.
- The optional-stack matrix lane runs the same test file on unpinned \`3.11\`/\`3.12\` (patch drifts with the runner tool cache), so the authority precondition fails.
- The dedicated \`articulated-manufactured-authority\` job (pins \`3.11.15\` + authority lock) is the governed lane for the authority profile; the repo's own contract tests (\`tests/ci/test_articulated_manufactured_hybrid_ci_red.py\`) treat the two lanes as distinct.

## Acceptance criteria
- The test exercises the authority profile only when the governed environment is actually satisfied, and the rolling profile (declared purpose: non-authoritative compatibility) otherwise.
- Byte-determinism (native repeats byte-exact) is asserted in both environments.
- Strict comparison against the committed authority record remains authority-environment-only.
- \`optional-stack-check\` green on main; authority job unchanged.

dieterolson:
## Summary
- \`optional-stack-check\` is red on **main** (run 33750885075): \`tests/research/test_articulated_manufactured_solution.py::test_manufactured_solution_record_is_byte_deterministic\` fails with \`RuntimeError: authority requires exact Python patch 3.11.15\`.
- \`write_record()\` defaults to \`profile="authority"\`, which hard-requires exact CPython 3.11.15 + the hash-locked dependency set. The optional-stack matrix lanes run unpinned \`3.11\`/\`3.12\` (patch drifts with the runner tool cache), so the precondition legitimately cannot hold there.
- Per the repo's own governance contract (\`tests/ci/test_articulated_manufactured_hybrid_ci_red.py\`), the authority profile belongs to the dedicated \`articulated-manufactured-authority\` job (pins \`3.11.15\` + authority lock); the matrix lanes are rolling territory.

## Change
- Test probes \`validate_authority_environment()\` and selects the profile:
  - **Governed stack live** → authority profile: behavior unchanged, including the strict compare against the committed authority record and the \`publication_authority == "authoritative"\` assertion.
  - **Otherwise** (optional-stack lanes, local dev) → rolling profile, asserting native-repeat byte-determinism (the actual point of the test) plus the \`non_authoritative_compatibility_only\` record marker.
- The dedicated authority CI job is untouched; all 6 hybrid-CI contract tests still pass.

## Verification
- \`pytest tests/ci/test_articulated_manufactured_hybrid_ci_red.py\` → 6 passed.
- \`ruff check\` + \`ruff format --check\` clean.
- Research test self-skips without native Pinocchio (unchanged guard); CI lanes with Pinocchio installed will now exercise the rolling path instead of failing on interpreter-patch drift.

Fixes #9451


dieterolson:
Pushed the SPEC.md §12 changelog row (1.0.718) required by the spec-freshness gate (source files changed without a SPEC update). The test change itself is unchanged.

